PR #126, opened 2020-03-24
fix: consider when error.bulkRequestIndex = 0

zaki-yama:
It seems `KintoneAllRecordsError.extractErrorIndex()` doesn't work correctly if `error.bulkRequestIndex = 0`.


zaki-yama:
@koba04 
Thank you for your advice!
I added a test in 6cc0298 .

PR #149, opened 2020-04-01
feat: support OAuth token

### LINKED ISSUE #110: feat: Support OAuth access token
## Motivation

Kintone has OAuth access token support, but `rest-api-client` doesn't support yet.
So we want `rest-api-client` to accept OAuth access token.

ref. #64 

## Overview

This PR make `rest-api-token` to be able to accept OAuth token at its constructor.

```ts
// This interface hasn't fixed so we might change the interface
const client = new KintoneRestAPIClient({
  baseUrl: "https://example.cybozu.com",
  auth: {
    oauthToken: process.env.KINTONE_OAUTH_TOKEN
  }
});
```

koba04:
fixes #110


koba04:
I've used "oAuthToken" as the token name.
Which do you like better, "oAuthToken" or "oAuthAccessToken"?

PR #177, opened 2020-04-16
chore(rest-api-client): move api client script to examples

koba04:
We have `scripts/` in `rest-api-client` package.
It's ok, but it leads to bring unnecessary dependencies and settings into the repository.
We are using `lerna` and `yarn`, so we can move the scripts to another directory like `examples` easily.

We should refactor those scripts, but I don't include them in this PR.
I'll work on it later.


examples/rest-api-client-demo/README.md
```diff
@@ -0,0 +1,47 @@
...
+  - Required if you use API token authentication.
+- `KINTONE_OAUTH_TOKEN`
+  - An OAuth access token for Kintone apps.
+  - Required if you use OAuth authentication.
+- `KINTONE_DOMAIN`
+  - A domain name for your Kintone environment.
+  - Required if you upload scripts to your Kintone environment using `@kintone/customize-uploader`.
+  - This will be migrated to `KINTONE_BASE_URL` in the future.
+
+### Modifiy hard-coded IDs
```
zaki-yama, line 26:
```suggestion
### Modify hard-coded IDs
```

PR #182, opened 2020-04-17
chore: add npm start

koba04:
This PR is to add an npm-scripts to create `lib/*.js` files with a watch mode.
This is useful for type checking, and executing a script in `examples/rest-api-client-demo`.
Because when we use `@kintone/rest-api-client` in other packages, `lib/*.js`files are used, so we have to compile to use the latest version of `@kintone/rest-api-client`.


koba04:
#184 solves the problem better way, but I think having the `start` script is useful, so I'm still opening the PR.

PR #184, opened 2020-04-19
build: use paths compiler option to run scripts without compiling

koba04:
After #177  had been merged, we need to compile `rest-api-client` to run scripts in `rest-api-client-demo`.
In order to fix this, I've created a PR #182, but we still need to compile with this PR.
It's cumbersome, so I take another option, which is a way to use the `paths` compiler option.

In this PR, we can run scripts as `yarn script record getRecord` regardless of compiling `rest-api-client`, which is the same experience before #177.

### DIFF
--- a/examples/rest-api-client-demo/README.md
+++ b/examples/rest-api-client-demo/README.md
@@ -33,7 +33,7 @@ We have an app ID in `customize-manifest.json`, so you have to update the app ID
 ### Run a script with `ts-node`
 
 ```
-% npx ts-node src/run-node.ts record getRecord
+% yarn script record getRecord
 ```
 
 ### Upload scripts to your Kintone environment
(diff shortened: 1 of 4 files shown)


examples/rest-api-client-demo/README.md
```diff
@@ -33,7 +33,7 @@ We have an app ID in `customize-manifest.json`, so you have to update the app ID
 ### Run a script with `ts-node`
 
 ```
-% npx ts-node src/run-node.ts record getRecord
+% yarn script record getRecord
```
shisama, line 36:
```suggestion
% yarn run-script record getRecord
```

examples/rest-api-client-demo/package.json
```diff
@@ -21,6 +21,7 @@
     "lint": "run-p -l lint:*",
     "deploy": "rimraf scripts/dist && run-s webpack upload",
     "upload": "kintone-customize-uploader customize-manifest.json",
+    "script": "ts-node -r tsconfig-paths/register src/run-node.ts",
```
shisama, line 24:
```suggestion
    "run-script": "ts-node -r tsconfig-paths/register src/run-node.ts",
```

PR #214, opened 2020-05-05
feat: Support Node.js Native ES Modules

shisama:
Node.js has already supported ES Modules.
see https://nodejs.org/api/esm.html

However, Node.js ESM is still experimental.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,5 +1,5 @@
 ---
-name: ":bug: Bug"
+name: "🐛 Bug"
 about: Issue for a bug
 title: ''
 labels: ''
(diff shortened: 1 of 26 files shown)


koba04:
@shisama 
Why did you add `examples/rest-api-client-demo/src/run-node.js` and `examples/rest-api-client-demo/src/run-node.mjs`?

shisama:
@koba04 
I added them to check easily if `rest-api-client` can be used from CJS/ESM in Node.js environment.

shisama:
I made this to draft because I am wondering whether I should [use rollup](https://github.com/kintone/js-sdk/pull/214/commits/24edb81c2c69acdb6d416dfb165c442b39aec360) or [add index.mjs using `module.createRequire](https://github.com/kintone/js-sdk/pull/214/commits/acc418fb1e46cee0cf8834525ce23359a5c22eaa) for Node.js native ESM.

Issue #232, opened 2020-05-15
Feature request: fileUpload supports Stream


TakashiAihara:
kintone-js-sdk のように、file.data で Streamを受け取れるようにしていただけると嬉しいです。

shisama:
@TakashiAihara Thank you for your proposal. We will release this as a new feature in a few months.

koba04:
We've added the Stream support in the documentation.
https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/docs/file.md#parameters

PR #234, opened 2020-05-19
docs: FileClient.uploadFile supports stream

chick-p:
Fixes: #232

`FileClient.uploadFile()` has already supported Stream because the method uses [form-data](https://www.npmjs.com/package/form-data) in Node.js environment.


koba04:
@chick-p `form-data` is only used in Node.js environment, so Stream support is only in Node.js. It would be nice to add a note for the limitation👍

chick-p:
@koba04 Thank you for your review. I added a note for the limitation.

PR #241, opened 2020-05-23
feat: support `garoon.connect.kintone.getRequestToken()` of Garoon JavaScript API

gorohash:
I would like to use this SDK in Garoon JavaScript customize.
But now, it's not supported to send request with request token getting by Garoon JavaScript API.

So I made this PR to support it.
Please review and merge!!!

cf. [kintone連携用トークンを取得する - cybozu developer network](https://developer.cybozu.io/hc/ja/articles/115005294843-kintone%E9%80%A3%E6%90%BA%E7%94%A8%E3%83%88%E3%83%BC%E3%82%AF%E3%83%B3%E3%82%92%E5%8F%96%E5%BE%97%E3%81%99%E3%82%8B)

### DIFF
--- a/packages/rest-api-client/src/KintoneRequestConfigBuilder.ts
+++ b/packages/rest-api-client/src/KintoneRequestConfigBuilder.ts
@@ -62,7 +62,7 @@ export class KintoneRequestConfigBuilder implements RequestConfigBuilder {
     this.proxy = proxy;
   }
 
-  public build(
+  public async build(
     method: HttpMethod,
     path: string,
     params: Data,
@@ -87,7 +87,7 @@ export class KintoneRequestConfigBuilder implements RequestConfigBuilder {
             ...requestConfig,
             method: "post" as const,
             headers: { ...this.headers, "X-HTTP-Method-Override": "GET" },
-            data: this.buildData(params),
+            data: await this.buildData(params),
           };
         }
         return {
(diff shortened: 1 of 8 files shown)


packages/rest-api-client/src/KintoneRequestConfigBuilder.ts
```diff
@@ -60,9 +62,10 @@ export class KintoneRequestConfigBuilder implements RequestConfigBuilder {
     this.headers = this.buildHeaders(basicAuth);
     this.clientCertAuth = clientCertAuth;
     this.proxy = proxy;
+    this.requestToken = "";
```
Kyome22, line 65: Please change the initial value of requestToken to null.
We don't want the empty string to have a special meaning.

gorohash: It's also possible to be undefined the initial value implicitly.
Is it better to be defined explicitly the initial value of requestToken as null?

shisama: Either `this.requestToken;` or `this.requestToken = null;` is OK. But we prefer to set the initial value explicitly with `null`.

shisama:
@gorohash 

Thank you for your efforts. We added the description of this feature into README.

packages/rest-api-client/src/KintoneRequestConfigBuilder.ts
gorohash: I fixed it!

koba04:
@gorohash  We've just released [v1.4.0](https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/CHANGELOG.md#140-2020-06-02)! Thank you!

PR #269, opened 2020-06-19
deps: update @cybozu/eslint-config to v11

koba04:
## Why

We want to use the latest version of `@cybozu/eslint-config`.

## What

I've updated `@cybozu/eslint-config` version to v11, which is the latest version.

## How to test

`yarn lint`

## Checklist


koba04:
We have to fix errors caused by `@typescript-eslint/ban-types`.

koba04:
I've fixed errors for `customize-uploader`, but it would take more time to fix errors for `rest-api-client` because it uses `object` and `{}` types heavily....

koba04:
This is stale, so I re-take this as #359

PR #271, opened 2020-06-19
refactor: add buildBaseUrl as a platform deps

koba04:
## Why

RestAPIClient gets a `baseUrl` from `location.origin` in a browser environment if `baseUrl` is not specified, which is a function depends on a browser environment.

## What

I've extracted the function as a platform dependency.

## How to test

`yarn lint` && `yarn test`

## Checklist

### DIFF
--- a/packages/rest-api-client/src/KintoneRestAPIClient.ts
+++ b/packages/rest-api-client/src/KintoneRestAPIClient.ts
@@ -89,19 +89,6 @@ export const errorResponseHandler = (
   throw new KintoneRestAPIError({ data, ...rest });
 };
 
-// asserts syntax fails with arrow functions.
-// see https://github.com/microsoft/TypeScript/issues/33602
-type AssertsOptions = (
-  options: Options
-) => asserts options is Options & {
-  baseUrl: string;
-};
-const assertsOptions: AssertsOptions = (options) => {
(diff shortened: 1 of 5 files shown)


packages/rest-api-client/src/KintoneRestAPIClient.ts
```diff
@@ -132,12 +119,12 @@ export class KintoneRestAPIClient {
   private baseUrl?: string;
 
   constructor(options: Options = {}) {
-    options.baseUrl = this.baseUrl = options.baseUrl ?? location?.origin;
```
koba04, line 135: Mutating an argument object should be avoided.

packages/rest-api-client/src/__tests__/KintoneRestAPIClient.test.ts
```diff
@@ -4,38 +4,36 @@ import {
...
       const baseUrl = "https://example.com";
-      beforeEach(() => {
-        injectPlatformDeps(browserDeps);
-      });
-
       it("should use location.origin in browser environment if baseUrl param is not specified", () => {
-        global.location = {
-          origin: "https://example.com",
-        };
+        injectPlatformDeps(browserDeps);
```
koba04, line 31: The tests depended on if `location.origin` exists, which is not a way we judge which platform we are in.

PR #273, opened 2020-06-19
fix: use location.{protocol, host} instead of location.origin

koba04:
## Why

IE11 doesn't provide `location.origin` in some cases.

https://stackoverflow.com/questions/32722032/how-does-ie11-populate-window-location-origin

## What

use `location.protocol` and `location.host` instead of `location.origin`

## How to test

`yarn test` and `yarn lint`

## Checklist


koba04:
This PR is based on #271 so this will be ready after that has been merged.

Issue #67, opened 2020-02-25
feat: support `x-cybozu-warning` header


zaki-yama:
ref. #65 

When retrieving records with query, 

 https://developer.cybozu.io/hc/ja/articles/202331474-%E3%83%AC%E3%82%B3%E3%83%BC%E3%83%89%E3%81%AE%E5%8F%96%E5%BE%97-GET-

> クエリ検索結果が10万件以上存在するとき、絞り込みが中断されます。
> その場合レスポンスヘッダに「x-cybozu-warning : Filter aborted because of too many search results.」が追加されます。

We need to support this header or other ways for users to know this warning.

koba04:
We've added an option named `featureFlags.enableAbortSearchError` into `KintoneRestAPIClient` to throw an error if a response includes `x-cybozu-warning : Filter aborted because of too many search results` as a response header.

The default value of the option is `false`, but we have a plan to make the default option `true` in the future major release.

PR #304, opened 2020-07-10
feat: Add a new option to throw an error for x-cybozu-warning

koba04:
## Why

Currently, we don't have a way to handle the `x-cybozu-warning` header.

ref. #67

## What

This PR adds a new option to handle the `x-cybozu-warning` header in the constructor of `KintoneRestAPIClient`.
When the option is enabled, `KintoneRestAPIClient` throws an error if a received response has the header with the message, "Filter aborted because of too many search results."

I have not decided on the name of the option yet.

Why does `KintoneRestAPIClient` throw an error instead of a warning?
Because the search results are aborted, so the response is not what we expect. We've considered it, but I didn't find out a valid use-case using the incomplete results.

**We have a plan to make this a default option at the next major release.**

## Checklist


koba04:
I'm considering use `console.warn` even if the option is disabled.

packages/rest-api-client/README.md
```diff
@@ -79,24 +79,26 @@ client.record
...
+| featureFlags                                |                              Object                              |                             | Feature flags that you can configure                                                                                                                                                                                                                                             |
+| featureFlags.enableAbortedSearchResultError |                             boolean                              |                             | An option whether throwing `KintoneAbortedSearchResultError` or not. `false` is the default value.                                                                                                                                                                               |
```
b4h0-c4t, line 101:
```suggestion
| featureFlags.enableAbortedSearchResultError |                             boolean                              |                             | An option whether to throw `KintoneAbortedSearchResultError` or not. `false` is the default value.                                                                                                                                                                               |
```

PR #311, opened 2020-07-14
fix: fix a order to check kintone/garoon object

chick-p:
## Why

An error occurred while attempting to request except `GET` in a Garoon environment.
`kintone` object does not exist in a Garoon environment, so throws `ReferenceError`.

## What

The first thing, check a type of `kintone` (and `garoon`)  whether its type is `object`.

## How to test

`yarn run lint` && `yarn test`

## Checklist

### DIFF
--- a/packages/rest-api-client/src/platform/__tests__/browser.test.ts
+++ b/packages/rest-api-client/src/platform/__tests__/browser.test.ts
@@ -1,0 +1,37 @@
+import { getRequestToken } from "../browser";
+
+describe("getRequestToken()", () => {
+  let originalKintone: any;
+  let originalGaroon: any;
+  beforeEach(() => {
+    originalGaroon = global.garoon;
+    originalKintone = global.kintone;
+  });
+  afterEach(() => {
+    global.garoon = originalGaroon;
+    global.kintone = originalKintone;
+  });
+  it("should get a request token in kintone", async () => {
+    global.kintone = {
+      getRequestToken: () => "dummy request token from kintone",
+    };
+    const requestToken = await getRequestToken();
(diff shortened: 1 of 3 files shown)


packages/rest-api-client/src/platform/__tests__/browser.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    originalGaroon = global.garoon;
+    originalKintone = global.kintone;
+  });
+  afterEach(() => {
+    global.garoon = originalGaroon;
+    global.kintone = originalKintone;
+  });
+  it("should get a request token in kintone", async () => {
+    global.kintone = {
+      getRequestToken: () => "dummy request token from kintone",
```
shisama, line 16: We prefer extract this `"dummy request token from kintone"` to a variable. What we want to verify is “requestToken is equal to the return value of `kintone.getRequestToken()`. So, extracting it to a variable makes the purpose of this test clearer.

packages/rest-api-client/src/platform/__tests__/browser.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    };
+    const requestToken = await getRequestToken();
+    expect(requestToken).toBe("dummy request token from kintone");
+  });
+  it("should get a request token in garoon", async () => {
+    global.garoon = {
+      connect: {
+        kintone: {
+          getRequestToken: async () => "dummy request token from garoon",
```
shisama, line 25: We prefer extract this `"dummy request token from garoon"` to a variable, too.

chick-p:
@shisama 
Thank you for your review comment.
I fixed the code so that extract to a variable.

PR #341, opened 2020-08-04
Attach type for record arguments.

b4h0-c4t:
## Why
`record` arguments need `Record` type.

## What
Can use `Record` type in calling `RecordClient`.

## How to test
Modify existing `RecordClient` type test to `Record`.

## Checklist

### DIFF
--- a/packages/rest-api-client/src/client/RecordClient.ts
+++ b/packages/rest-api-client/src/client/RecordClient.ts
@@ -17,37 +17,9 @@ const ADD_RECORDS_LIMIT = 100;
 const UPDATE_RECORDS_LIMIT = 100;
 const DELETE_RECORDS_LIMIT = 100;
 
-type EntityForParameter = { code: string };
-type FileInformationForParameter = { fileKey: string };
-type SubtableFieldValue =
-  | number
-  | string
-  | string[]
-  | null
-  | FileInformationForParameter[]
-  | EntityForParameter[];
-type SubtableForParameter = {
-  [fieldCode: string]: Array<
-    | {
-        value: SubtableFieldValue;
-      }
-    | {
-        id: number | string;
-        value?: SubtableFieldValue;
-      }
-  >;
-};
 type RecordForParameter = {
   [fieldCode: string]: {
-    value:
-      | number
-      | string
-      | string[]
-      | null
-      | FileInformationForParameter[]
-      | EntityForParameter
-      | EntityForParameter[]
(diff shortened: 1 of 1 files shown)


packages/rest-api-client/src/client/RecordClient.ts
```diff
@@ -17,6 +17,32 @@ const ADD_RECORDS_LIMIT = 100;
 const UPDATE_RECORDS_LIMIT = 100;
 const DELETE_RECORDS_LIMIT = 100;
 
+type Entity = { code: string; name: string };
```
shintaroNagata, line 20: In parameter, `name` property is cannot specified.

packages/rest-api-client/src/client/RecordClient.ts
```diff
@@ -17,6 +17,32 @@ const ADD_RECORDS_LIMIT = 100;
...
+      | {
+          [fieldCode: string]: {
+            value:
+              | number
+              | string
+              | string[]
+              | null
+              | FileInformation[]
+              | Entity[];
+          };
```
shintaroNagata, line 41: The `value` parameter of a Subtable field has Array type value, and their elements can have `id` property.
Also, `value` property can be optional when `id` property is specified.

packages/rest-api-client/src/client/RecordClient.ts
```diff
@@ -93,7 +130,9 @@ export class RecordClient {
     }
     return this.addRecord({
       app,
-      record: { ...record, [updateKey.field]: { value: updateKey.value } },
+      record: Object.assign(record, {
```
shisama, line 133:
```suggestion
      record: Object.assign({}, record, {
```

packages/rest-api-client/src/client/RecordClient.ts
```diff
@@ -17,35 +17,37 @@ const ADD_RECORDS_LIMIT = 100;
...
-              }
-            | {
-                id: number | string;
-                value?: SubtableFieldValue;
-              };
-        };
+      | FileInformationForParametor[]
+      | EntityForParametor
+      | EntityForParametor[]
+      | SubtableForParametor;
```
shisama, line 50: @b4h0-c4t I found some misspellings. Please look for others.

```suggestion
      | FileInformationForParameter[]
      | EntityForParameter
      | EntityForParameter[]
      | SubtableForParameter;
```

PR #342, opened 2020-08-04
refactor: attach type of Layout to method argument

chick-p:
## Why

checks type of Layout when the method argument takes `layout`.

## What

- attach `LayoutForParaemeter` type to the method's parameter.
- fix test and add layout variation.

## How to test

`yarn test`

## Checklist

### DIFF
--- a/packages/rest-api-client/src/client/AppClient.ts
+++ b/packages/rest-api-client/src/client/AppClient.ts
@@ -92,7 +92,7 @@ export class AppClient {
 
   public updateFormLayout(params: {
     app: AppID;
-    layout: object[];
+    layout: Layout;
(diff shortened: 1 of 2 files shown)


zaki-yama:
@chick-p 
Thank you!
Are there any reasons why you define types (`~LayoutForParameter`) directly in AppClient.ts instead of creating `src/client/types/app/layout.ts` ?

chick-p:
@zaki-yama 
Thank you for reviewing and your comment!

I have defined type to same `AppClient.ts` as a first step.
We similarly have plans to attach type to the argument of other methods. (ref. #341)

I'd like to extract type definition files as another pull request if necessary.

PR #353, opened 2020-08-13
docs: fix syntax error for Overview code in file.md of rest-api-client

shisama:
## Why

Example code for Overview in file.md occurs syntax error.

## What

Added `}` and format it.

## How to test

Copy, paste, and execute it in your environment!!

## Checklist

### DIFF
--- a/packages/rest-api-client/docs/file.md
+++ b/packages/rest-api-client/docs/file.md
@@ -35,7 +35,7 @@ const client = new KintoneRestAPIClient();
     app: APP_ID,
     id,
   });
-  const data = await this.client.file.downloadFile({
+  const data = await client.file.downloadFile({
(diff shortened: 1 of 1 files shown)


packages/rest-api-client/docs/file.md
```diff
@@ -35,7 +36,7 @@ const client = new KintoneRestAPIClient();
     id,
   });
   const data = await this.client.file.downloadFile({
```
zaki-yama, line 38: I might have found another mistake. is `this` here unnecessary?
Would you remove it as well?

zaki-yama:
@shisama 
Thank you for fixing!
You're right... and I found another mistake while testing the script.
Please check.

packages/rest-api-client/docs/file.md
shisama, line 38:
```suggestion
  const data = await client.file.downloadFile({
```

PR #380, opened 2020-08-27
fix: KintoneRestAPIClient.version differs between UMD files and npm

shisama:
## Why

Using rest-api-client with UMD build file, KintoneRestAPIClient.version shows old version.

For example, When we use @kintone/rest-api-client@1.5.0, UMD file shows `1.4.2` running `KintoneRestAPIClient.version`.
see https://unpkg.com/@kintone/rest-api-client@1.5.0/umd/KintoneRestAPIClient.js

## What

The reason why we make such a mistake is below.
KintoneRestAPIClient.version refers to packageJson.version. Actually, we inject the version into UMD file with webpack. https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/webpack.config.js#L33
We update version in package.json and publish the package after building UMD file with webpack.

The solution is just adding 'prepublish' command into rest-api-client/package.json.
We need to build UMD file after updating version in package.json.
Lerna has a lifecycle in 'publish' command. 
https://github.com/lerna/lerna/tree/master/commands/publish#lifecycle-scripts
We can use lifecycle-scripts for each package.

## How to test

Run 'yarn release'

## Checklist

### DIFF
--- a/packages/rest-api-client/package.json
+++ b/packages/rest-api-client/package.json
@@ -16,6 +16,7 @@
   "scripts": {
     "build": "npm-run-all -l -s clean -p build:*",
     "lint": "run-p -l lint:*",
+    "prepublish": "npm run build",
     "prerelease": "npm-run-all -p lint test -s build",
     "test": "jest",
     "test:ci": "jest --runInBand",
--- a/packages/rest-api-client/package.json
+++ b/packages/rest-api-client/package.json
@@ -17,7 +17,7 @@
     "build": "npm-run-all -l -s clean -p build:*",
     "lint": "run-p -l lint:*",
     "prepublish": "npm run build",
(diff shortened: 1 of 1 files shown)


packages/rest-api-client/package.json
```diff
@@ -16,7 +16,8 @@
   "scripts": {
     "build": "npm-run-all -l -s clean -p build:*",
     "lint": "run-p -l lint:*",
-    "prerelease": "npm-run-all -p lint test -s build",
+    "prepublish": "npm run build",
```
zaki-yama, line 19:
```suggestion
    "prepublishOnly": "npm run build",
```

zaki-yama (changes requested):

zaki-yama (approved):

koba04 (approved):
👍

PR #381, opened 2020-08-28
fix: `npm start` with the modern template

koba04:
## Why

We've released a new template called `modern` in the `create-plugin` package. 
But the `npm start` isn't succeed at the first time. The reason is that running `uploader` at the first time doesn't have target files created by `webpack` because the script run `webpack` and `uploader` parallel.

## What

I've added `prebuild` script to run `webpack` before `npm start`.

## How to test

```
% yarn build
% yarn create-plugin --template modern ~/target/dir
% cd ~/target/dir
% npm start
:
```

## Checklist

### DIFF
--- a/packages/create-plugin/templates/modern/package.json.tmpl
+++ b/packages/create-plugin/templates/modern/package.json.tmpl
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "scripts": {
     <% if (enablePluginUploader) { %>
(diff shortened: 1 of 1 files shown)


chick-p (approved):
Thank you for fix it!
I try to execute and confirmed to be able to upload a plugin.

koba04:
@chick-p Thank you for your review!

PR #387, opened 2020-08-28
feat: support Node.js native ESM for rest-api-client

shisama:
## Why

Node.js has already supported ES Modules.
see https://nodejs.org/api/esm.html

However, Node.js ESM is still experimental.

We can import `@kintone/rest-api-client` in a ESM file now.
But we cannot use named import because the package provides only CJS file. So, we have to write code like the below.

```js
import mod from "@kintone/rest-api-client";
const { KintoneRestAPIClient } = mod;
```

I think it is more useful to use named import like below.

```js
import { KintoneRestAPIClient } from "@kintone/rest-api-client";
```

## What

- [Conditional Exports](https://nodejs.org/api/esm.html#esm_conditional_exports) is used to provide ESM file. I add "exports" field into package.json.
- Bundle files to resolve import path. All import paths in TypeScript source files don't have extensions. ESM needs file extensions in import paths.
- `require` cannot be used in ESM file. I added packageJson.ts to be available to inject package.json object from outside. So we can inject package.json object from ESM/CJS entry files.

## Checklist

### DIFF
--- a/packages/rest-api-client/package.json
+++ b/packages/rest-api-client/package.json
@@ -22,6 +22,7 @@
     "test:ci": "jest --runInBand",
     "build:umd_dev": "rollup -c --environment BUILD:development",
     "build:umd_prod": "rollup -c --environment BUILD:production",
+    "build:mjs": "rollup --config rollup.config.esm.js",
     "build:compile": "run-p compile",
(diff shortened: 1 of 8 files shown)


shisama:
As #436 is preferred, this closes.

PR #418, opened 2020-09-14
fix: kintone-rest-api-client-demo requires ts-loader at deploy

chick-p:
## Why

In `examples/kintone-rest-api-client-demo`

In webpack.config.js, `webpack` uses ts-loader but `ts-loader` does not exist in dependency.
ref: https://github.com/kintone/js-sdk/blob/master/examples/rest-api-client-demo/webpack.config.js#L21

Therefore, execute `yarn deploy`, then fail to build customized files.

## What

add `ts-loader` to dependency

## How to test

`yarn deploy`

## Checklist


koba04 (approved):
👍  Thanks!

PR #404, opened 2020-09-09
feat: export type definitions for Kintone fields from package root

b4h0-c4t:
## Why
KintoneFieldsType is required for the most packages.

## What
Export KintoneFields form `@kintone/rest-api-client`.

## How to test
none.

## Checklist

- [x] Read [CONTRIBUTING.md](https://github.com/kintone/js-sdk/blob/master/CONTRIBUTING.md)
- [x] Updated documentation if it is required.
- [x] Added tests if it is required.
- [x] Passed `yarn lint` and `yarn test` on the root directory.

PR #442, opened 2020-09-29
Supporting types generated dts gen

mura-:
## Why

Index signature type error occurs when I using types generated by kintone-dts-gen as generics. 
image

I want to use types generated by kintone-dts-gen without casting.

## What

I've extended `Record` type to support kintone-dts-gen types as well.
Could you please check if this implementation is okay?

## How to test

Now I editing.

## Checklist

### DIFF
--- a/packages/rest-api-client/src/client/RecordClient.ts
+++ b/packages/rest-api-client/src/client/RecordClient.ts
@@ -50,16 +50,6 @@ export class RecordClient {
     return this.client.get(path, params);
   }
 
-  public getRecord2<T>(params: {
-    app: AppID;
-    id: RecordID;
-  }): Promise<{ record: Record<T> }> {
-    const path = this.buildPathWithGuestSpaceId({
-      endpointName: "record",
-    });
-    return this.client.get(path, params);
-  }
-
   public addRecord(params: {
     app: AppID;
     record?: RecordForParameter;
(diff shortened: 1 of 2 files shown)


koba04:
@mura- 
Thank you for submitting a pull request!
We've recognized the issue.

Extending the `Record` is not a way we intent.
Unfortunately, the type definition `dts-gen` generates is intended to use with kintone JS API rather than `rest-api-client`, so it doesn't work well with `rest-api-client`.
But we think it is not ideal, so we are going to provide a way to generate type definitions that work well with `rest-api-client` in the future roadmap.

We are currently working on exposing type definitions to define type definitions for your app.
It makes it possible to define type definitions for your application, it requires defining type definitions manually, though.
https://github.com/kintone/js-sdk/pull/404

mura-:
@koba04 

> Extending the Record is not a way we intent.

I understood. 👍

> We are currently working on exposing type definitions to define type definitions for your app.
> It makes it possible to define type definitions for your application, it requires defining type definitions manually, though.

Got it. 
I'm looking forward to the completion of that PR 🙌

koba04:
@mura- Thank you for your understanding! This is very helpful feedback for us, thank you!

koba04:
@mura- 
I've created an issue to fix this as #445 
So I'll close this as I mentioned at https://github.com/kintone/js-sdk/pull/442#issuecomment-701181764, but thank you for your work!!! 👏 👏 👏

PR #469, opened 2020-10-15
fix: Modify `properties` type in AppClient

b4h0-c4t:
## Why
AppClient needs properties type for parameter.
Because as it stands now, the parameter isn't being suggested properties.

## What
Add type definition for properties as the parameter.

As a trade-off for suggesting a property, you need to specify "as const" for the property "types"

example
```ts
const properties = {
  fieldCode: {
    type: "SINGLE_LINE_TEXT" as const,
    code: "fieldCode",
    label: "Text Field",
  },
};

appClient.updateFormFields({ app, properties, revision });
```

## How to test
Existing

## Checklist

### DIFF
--- a/examples/rest-api-client-demo/src/app.ts
+++ b/examples/rest-api-client-demo/src/app.ts
@@ -30,7 +30,7 @@ export class App {
   public async addFormFields() {
     const properties = {
       fieldCode: {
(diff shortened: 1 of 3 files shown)


zaki-yama (approved):
👍

PR #474, opened 2020-10-16
refactor: move type definitions for auth

shintaroNagata:
## Why

Type definitions for authentication object are located in KintoneRestAPIClient.ts.
However, since these definitions are independent from the definition of `KintoneRestAPIClient`,  they should be located in another file.
Actually, a circular reference occurs between KintoneRestAPIClient.ts and KintoneRequestConfigBuilder.ts because `KintoneRequestConfigBuilder` refers them.

## What

I create a new file to locate them & move them to there.

## How to test

`yarn test` & `yarn lint`

## Checklist


koba04 (approved):
Great, LTGM 👍

PR #506, opened 2020-10-30
fix: adjust browserslist to product

chick-p:
## Why

`rest-api-client` supports any browsers.
However, `browserslist` of PC have not matched [supported web browsers in Kintone](https://get.kintone.help/general/en/user/list_start/webbrowser.html).

## What

- match the description in `README.md` to the supported web browsers
- fix build settings for iOS as supporting above version 11

## How to test

`yarn build` and `yarn lint`

## Checklist


packages/rest-api-client/rollup.config.js
```diff
@@ -34,10 +34,10 @@ export default {
...
                 "IE 11",
-                "last 2 edge versions",
-                "last 2 firefox version",
-                "last 2 chrome versions",
-                "last 2 safari versions",
+                "last 1 edge versions",
+                "last 1 firefox version",
+                "last 1 chrome versions",
+                "last 1 safari versions",
                 "iOS 11",
```
b4h0-c4t, line 41:
```suggestion
                "iOS 12",
```

reference in [Kintone Supported Web Browsers](https://get.kintone.help/general/en/user/list_start/webbrowser.html)

packages/rest-api-client/rollup.config.js
```diff
@@ -34,10 +34,10 @@ export default {
...
                 "last 2 and_chr versions",
```
b4h0-c4t, line 42:
```suggestion
                "last 1 and_chr versions",
```

reference in [Kintone Supported Web Browsers](https://get.kintone.help/general/en/user/list_start/webbrowser.html)

b4h0-c4t (commented):
There were several unchanged lists.

packages/rest-api-client/rollup.config.js
chick-p: Thank you for your suggestion!
Because select the iOS browser version 12 directly, I fixed iOS browser version as `iOS >= 12`.

b4h0-c4t:
The supported browsers version is dynamic, it is understood that you should use the "latest 2 versions" to avoid unexpected behavior.

On the other hand, can you tell me why you don't set the iOS versions to 12 or higher?

b4h0-c4t (commented):
The supported browsers version is dynamic, it is understood that you should use the "latest 2 versions" to avoid unexpected behavior.

On the other hand, can you tell me why you don't set the iOS versions to 12 or higher?

chick-p:
@b4h0-c4t 
Thanks.
In `master`, the iOS version is 11.
If it intendedly changes 12 or higher, the injected polyfill may change, and it may be BREAKING CHANGE.
This PR is that makes the README match browsers support of the product, so  I think which there is no need for this PR.

b4h0-c4t (approved):

Issue #530, opened 2020-11-11
Unify environment variables used in js-sdk packages 


koba04:
## Tasks

- [x] Add missing environment variables and CLI flags support in `@kintone/plugin-uploader`  @chick-p 
  - [x] `KINTONE_BASE_URL` #534
- [x] Add missing environment variables  and CLI flags support in `@kintone/customize-uploader` @chick-p 
  - [x] `KINTONE_BASE_URL` #543
  - [x] `KINTONE_OAUTH_TOKEN` #566
- [x] Add missing environment variables  and CLI flags support in `@kintone/dts-gen` @koba04 (@potaracom)
  - [x] `KINTONE_USERNAME` #532 (#458)
  - [x] `KINTONE_PASSWORD` #532 (#458)
  - [x] `KINTONE_BASE_URL` #536
  - [x] `KINTONE_API_TOKEN` #575
  - [x] `KINTONE_OAUTH_TOKEN` #575 
  - [x] `KINTONE_BASIC_AUTH_USERNAME` #532 (#458)
  - [x] `KINTONE_BASIC_AUTH_PASSWORD ` #532 (#458)
  - [x] `HTTP_PROXY` #585 
  - [x] `HTTPS_PROXY ` #585 
- [x] Release a minor version of the packages
- [x] Add deprecated warnings for the usage of `KINTONE_DOMAIN`, which includes `domain` option as well @chick-p 
  - [x] `@kintone/plugin-uploader` #697 
  - [x] `@kintone/customize-uploader` #602 
- [x] Add deprecated warnings for the usage of `host`, `proxyHost`, and `proxyPort` option
  - [x] `@kintone/dts-gen` #869
- [x] Release a minor version of the packages
  - [`@kintone/dts-gen@4.0.0`](https://github.com/kintone/js-sdk/blob/master/packages/dts-gen/CHANGELOG.md#400-2021-05-11)
  - [`@kintone/plugin-uploader@4.3.0`](https://github.com/kintone/js-sdk/blob/master/packages/plugin-uploader/CHANGELOG.md#430-2021-03-09)
  - `@kintone/customize-uploader@3.1.1` (No CHANGELOG)
- [x]  Remove deprecated environment variables and CLI flags from the packages.
  - [x] `@kintone/plugin-uploader` #1046
  - [x] `@kintone/customize-uploader` #1047 
  - [x] `@kintone/dts-gen` #1048
- [x]  Release a major version of the packages :tada:

## Package

[...]

PR #534, opened 2020-11-15
feat: support environment variable for base URL of plugin-uploader

chick-p:
## Why

This is a part of #530

## What

I've added an environment variable and a CLI flag support in plugin-uploader.

- `KINTONE_BASE_URL`

## How to test

- set `KINTONE_BASE_URL`to environment valiable and execute `bin/cli.js SAMPLE_PLUGIN.zip`
- execute `bin/cli.js SAMPLE_PLUGIN.zip` with `--base-url` option

## Checklist

### DIFF
--- a/packages/plugin-uploader/README.md
+++ b/packages/plugin-uploader/README.md
@@ -47,6 +47,7 @@ You can create a project based on `@kintone/plugin-packer` using [@kintone/creat
     $ kintone-plugin-uploader <pluginPath>
   Options
     --domain Domain of your kintone
+    --base-url Base-url of your kintone (If you set domain, its value is not necessary.)
     --username Login username
     --password User's password
     --proxy Proxy server
@@ -57,6 +58,7 @@ You can create a project based on `@kintone/plugin-packer` using [@kintone/creat
 
     You can set the values through environment variables
     domain: KINTONE_DOMAIN
+    base-url: KINTONE_BASE_URL (If you set domain, its value is not necessary.)
     username: KINTONE_USERNAME
     password: KINTONE_PASSWORD
     basic-auth-username: KINTONE_BASIC_AUTH_USERNAME
--- a/packages/plugin-uploader/README.md
+++ b/packages/plugin-uploader/README.md
@@ -46,8 +46,8 @@ You can create a project based on `@kintone/plugin-packer` using [@kintone/creat
   Usage
     $ kintone-plugin-uploader <pluginPath>
   Options
-    --domain Domain of your kintone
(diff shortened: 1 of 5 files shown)


packages/plugin-uploader/README.md
```diff
@@ -47,6 +47,7 @@ You can create a project based on `@kintone/plugin-packer` using [@kintone/creat
     $ kintone-plugin-uploader <pluginPath>
   Options
     --domain Domain of your kintone
+    --base-url Base-url of your kintone (If you set domain, its value is not necessary.)
```
shisama, line 50: I think `--base-url` is preferred to `--domain` because we will delete `--domain` in the future. So it seems to be better to introduce `--base-url` earlier than `--domain`.

```suggestion
    --base-url Base-url of your kintone
    --domain Domain of your kintone (If you set `--base-url`, this value is not necessary.)
```

packages/plugin-uploader/src/params.ts
```diff
@@ -6,18 +6,25 @@ interface Params {
...
   const questions: inquirer.Question[] = [
     {
       type: "input",
       message: m("Q_Domain"),
       name: "domain",
       default: domain,
-      when: () => !domain,
+      when: () => !baseUrl && !domain,
       validate: (v: string) => !!v,
     },
```
shisama, line 29: Will you please tell me the reason not to add `Q_BaseUrl` into `src/message.ts`?

chick-p: I think that `Q_BaseUrl`  need is when adding `baseUrl` question into `questions`
Even if add `baseUrl`, it can only ask one or the other about kintone domain.

shisama: Are you gonna add `Q_BaseUrl` and `baseUrl` question when we deprecate `KINTONE_DOMAIN`? If so, `Q_Domain` and `domain` question will be removed, won't it?

chick-p: Yes, I guess the addition timing is when it is deprecated `KINTONE_DOMAIN`.
Would it be better to add the question of baseUrl in this PR, even if not called until deprecating `KINTONE_DOMAIN`?

chick-p: I have added the base-url question as [b084014](https://github.com/kintone/js-sdk/pull/534/commits/b08401463400ec85c0ac85b6f0a7d56b16ac255f).

shisama (approved):
@chick-p LGTM 👍 Thank you for your work!

PR #536, opened 2020-11-18
feat: support --base-url and KINTONE_BASE_URL

koba04:
## Why

This is a part of #530.

## What

I've added `--base-url` option and support the `KINTONE_BASE_URL` environment value.
I've also replaced `createCommanderProgram` with `cli-parser` to encapsulate the logic for parsing the `process.argv` into `cli-parser`.

## Checklist


packages/dts-gen/src/cli-parser-test.ts
```diff
@@ -0,0 +1,134 @@
+import { parse } from "./cli-parser";
```
shisama, line 1: If this should run when `yarn test` is executed, `cli-parser.test.ts` is the correct file name. Now, this file doesn't run when `yarn test` is executed.

packages/dts-gen/src/cli-parser-test.ts
```diff
@@ -0,0 +1,134 @@
...
+            );
+            expect(args.username).toBe("USERNAME");
+            expect(args.password).toBe("PASSWORD");
+            expect(args.appId).toBe("APP_ID");
+            expect(args.guestSpaceId).toBe(
+                "GUEST_SPACE_ID"
+            );
+            expect(args.preview).toBe(true);
+            expect(args.namespace).toBe("NAMESPACE");
+            expect(args.proxyHost).toBe("PROXY_HOST");
```
shisama, line 94: Maybe `typeName` is forgot.

packages/dts-gen/src/cli-parser-test.ts
```diff
@@ -0,0 +1,134 @@
...
+        });
+    });
+    describe("validations", () => {
+        test("unspecified baseUrl", () => {
+            expect(() => {
+                parse([
+                    "node",
+                    "index.js",
+                    "--base-url",
+                    "https://example2.kintone.com",
```
shisama, line 127: This test fails since `baseUrl` is specified. This URL literal may not need.
```suggestion
```

packages/dts-gen/src/cli-parser.ts
```diff
@@ -72,5 +98,17 @@ export function createCommanderProgram() {
...
+    if (baseUrl === null) {
+        throw new Error(
+            "--base-url (KINTONE_BASE_URL) must be specified"
+        );
+    }
+
+    return {
+        ...parsedArgs,
+        baseUrl,
+    } as any;
```
shisama, line 113:
```suggestion
    } as ParsedArgs;
```

shisama (changes requested):
@koba04 Good job! Thank you for fixing this. I left some comments. Please take a look at them.

packages/dts-gen/src/cli-parser.ts
koba04: 👍  Unfortunately, the type cannot be casted to ParsedArgs directly because the type of `parsedArgs` is `commander.Command`, so I've casted the type at the `program.parse()`.
https://github.com/kintone/js-sdk/pull/536/commits/3d3f6bb541d2fb303f0da5ccd45c3e007684d7b6#diff-75257f38fd4cf6ea78b0699d85fb15d6e6a998407380ff314282ce47948ae580R101-R104

koba04:
@shisama Thank you for your review 🌟  I've addressed the points you've pointed out!

packages/dts-gen/src/cli-parser.ts
shisama: Oh, sorry. I've tried this now on my local and I realized this is wrong.

shisama: BTW, the `parsedArgs` variable has properties that `ParsedArgs` does not have. For example, `host`. Is it OK to destructure the object (`...parsedArgs`)?

koba04: That's a good point. I've fixed it. 👍

shisama (approved):
@koba04 👍 Thank you for fixing what I pointed out.

koba04:
Thank for the good feedbacks!!! 🎉 🎉 🎉

PR #566, opened 2020-12-01
feat: support OAuth token of customize uploader

chick-p:
## Why

This is a part of #530

## What

- add `--oauth-token` option and support the `KINTONE_OAUTH_TOKEN` environment value.
- support authorization using an OAuth access token.

## How to test

1. specify the following scopes and generate OAuth access tokens:
    - for upload
      - k:file:write
      - k:app_settings:read
      - k:app_settings:write
    - for import
      - k:file:read
      - k:app_settings:read
1. build and run it.
    ```shell
    $ yarn
    $ cd packages/customize-uploader
    $ yarn build
    $ node bin/cli.js sample/customize-manifest.json --oauth-token YOUR_OAUTH_ACCESS_TOKEN_FOR_UPLOAD
    $ node bin/cli.js import sample/customize-manifest.json --oauth-token YOUR_OAUTH_ACCESS_TOKEN_FOR_IMPORT
    ```

## Checklist

### DIFF
--- a/packages/customize-uploader/README.md
+++ b/packages/customize-uploader/README.md
@@ -64,6 +64,7 @@ If you want to upload the customize files automatically when a file is updated, 
     --domain Domain of your kintone (If you set --base-url, this value is not necessary.)
     --username Login username
     --password User's password
+    --oauth-token OAuth access token (If you set a set of --username and --password, this value is not necessary.)
     --basic-auth-username Basic Authentication username
     --basic-auth-password Basic Authentication password
     --proxy Proxy server
@@ -84,6 +85,7 @@ If you want to upload the customize files automatically when a file is updated, 
     domain: KINTONE_DOMAIN (If you set `base-url`, this value is not necessary.)
     username: KINTONE_USERNAME
     password: KINTONE_PASSWORD
+    oauth-token: KINTONE_OAUTH_TOKEN (If you set a set of username and password, this value is not necessary.)
     basic-auth-username: KINTONE_BASIC_AUTH_USERNAME
     basic-auth-password: KINTONE_BASIC_AUTH_PASSWORD
     proxy: HTTPS_PROXY or HTTP_PROXY
--- a/packages/customize-uploader/README.md
+++ b/packages/customize-uploader/README.md
@@ -99,7 +99,7 @@ If you omit the options, you can input the options interactively.
 ? Input your password: [hidden]
 ```
 
-If you use OAuthToken, you need to grant scopes corresponding to the command.
+If you use OAuth access token, you need to grant scopes corresponding to the command.
 
 - To upload a customize setting:
   `k:app_settings:write` and `k:file:write`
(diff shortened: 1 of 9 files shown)


packages/customize-uploader/README.md
```diff
@@ -97,6 +99,15 @@ If you omit the options, you can input the options interactively.
 ? Input your password: [hidden]
 ```
 
+If you use OAuthToken, you need to grant scopes corresponding to the command.
```
zaki-yama, line 102: Is "OAuthToken" intended to a variable/parameter ?
If not, I prefer to write "OAuth access token" (as you write in the description of `--oauth-token`). 👍 
Or just "If you use `--oauth-token`, ..."

zaki-yama (commented):
@chick-p 
thank you!
I left a comment, but it's a small thing.

packages/customize-uploader/README.md
chick-p: I've used "OAuthToken" in the sense of a common noun.
I will change to "OAuth access token" what you told me.

chick-p: fixed in cf8f649

chick-p:
@zaki-yama 
 Thank you for your review! I fixed based on your comment.
Additionally, I fix to mistake required scope when uploading a customized setting.

packages/customize-uploader/src/commands/__tests__/import.test.ts
```diff
@@ -32,6 +32,7 @@ describe("import", () => {
       kintoneApiClient = new MockKintoneApiClient(
         "kintone",
         "hogehoge",
+        "oauthToken",
```
zaki-yama, line 35:
```suggestion
        "oAuthToken",
```

packages/customize-uploader/src/params/index.ts
```diff
@@ -5,6 +5,7 @@ import { getBoundMessage } from "../messages";
 interface Params {
   username?: string;
   password?: string;
+  oauthToken?: string;
```
zaki-yama, line 8: I just wondered why you had remained this parameter `oauthToken`, not `oAuthToken`.
Is it also the variable we can't rename to `oAuthToken` due to the library?

zaki-yama (commented):
@chick-p 
Thanks!
I left a few comments more. Please check 🙏

packages/customize-uploader/src/params/index.ts
chick-p: No, the variable in `/src/params/index.ts` is not receiving restriction on the name the as `bin/cli.js`.
I thought that need to rename to `oAuthToken` after passing to the layer of commands.
But the restriction is given only by `meow`, I renamed it by what you say.

zaki-yama:
@chick-p 

> Additionally, I fix to mistake required scope when uploading a customized setting.

Thank you!
I edited the description of this PR, too.

zaki-yama (approved):
👍

koba04 (approved):
LTGM 👍 
I think `oAuthToken` will never be `null` because the default value is `""`, but it's another issue. Let's address it someday.

PR #585, opened 2020-12-08
feat: support HTTP(S)_PROXY and --proxys option  in dts-gen

koba04:
## Why

This is a part of #530.

## What

I've added `HTTPS_PROXY` and `HTTP_PROXY` support in `dts-gen`.
In addition, I've added the `--proxy` option.
`proxyHost` and `proxyPort` are inconsistent with other packages like `customize-uploader` and `plugin-uploader`.
I'll deprecate these command line options in the future.

## Checklist

### DIFF
--- a/packages/dts-gen/docs/how-to-use.md
+++ b/packages/dts-gen/docs/how-to-use.md
@@ -68,6 +68,23 @@ is defined in `sample-fields.d.ts`.
 
 **If you change form settings in kintone, Please re-generate type definition files**
 
+### Proxy support
+
+You can use a proxy server with `--proxy` option.
+
+```bash
+$ kintone-dts-gen --base-url https://***.cybozu.com \
+                 -u username \
+                 -p password \
+                 --proxy http://localhost:8000 \
+                 --app-id 12 \
+                 --type-name SampleFields \
+                 --namespace company.name.types \
+                 -o sample-fields.d.ts
+```
+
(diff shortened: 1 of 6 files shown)


chick-p (approved):
@koba04 
Thank you for your great work!
I confirmed that `--proxy` option and the environment variables is work correctly.

packages/dts-gen/docs/how-to-use.md
```diff
@@ -68,6 +68,23 @@ is defined in `sample-fields.d.ts`.
...
+                 -u username \
+                 -p password \
+                 --proxy http://localhost:8000 \
+                 --app-id 12 \
+                 --type-name SampleFields \
+                 --namespace company.name.types \
+                 -o sample-fields.d.ts
+```
+
+In addition, you can use `HTTP_PROXY` and `HTTPS_PROXY` environment variables instead of the command line option.
```
chick-p, line 86: dts-gen have been supported `--proxy-host` and `--proxy-port`.
Would you like to write about it?

koba04: @chick-p That's a good point. I'll deprecate the options (#530) because the options are not consistent with other packages.
In addition, the options are not useful than the `--proxy` option because authentication for a proxy server is not supported with the options as long as we add `--proxyUserName` and `--proxyPassword`.

chick-p: @koba04 
Thank you for explaining it in detail. I understood the reason!

PR #601, opened 2021-01-05
feat: export error classes to allow users to import them

shisama:
## TODO
- [x] add tests for error exported from index.ts
- [x] add documents to indicate how to import error classes

## Why

This allows users to import error classes like KintoneRestAPIError so that they can use them to check which error like the following code and to do other things.

```javascript
import { KintoneRestAPIClient, KintoneRestAPIError } from "@kintone/rest-api-client";

try {
  const client = new KintoneRestAPIClient(baseUrl, auth);
  const records = client.record.getRecord({ app, id });
  // do something with the records variables
} catch (e) {
  if (e instanceof KintoneRestAPIError) {
    // do something if the error is KintoneRestAPIError
  }
}
```

## What

Classes in `src/error` are exported as a named export in index.ts or some entry points.

## How to test

```
yarn test
```

## Checklist

### DIFF
--- a/packages/rest-api-client/docs/errorHandling.md
+++ b/packages/rest-api-client/docs/errorHandling.md
@@ -2,7 +2,18 @@
 
 - [KintoneRestAPIError](#kintonerestapierror)
 - [KintoneAllRecordsError](#kintoneallrecordserror)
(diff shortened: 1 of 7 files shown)


chick-p (approved):
@shisama 
Thank you for your quick work! 
LGTM👍

PR #604, opened 2021-01-12
feat(rest-api-client): implement GET /k/v1/app/notifications/general.json 

chick-p:
## Why

In this PR, support `GET /k/v1/app/notifications/general.json`.
SPEC: https://developer.cybozu.io/hc/ja/articles/900004873643

## What

- implement `getGeneralNotifications()` method in `AppClient`
- add a test
- add an example script
- write a documentation

## How to test

```shell
$ yarn install
$ yarn build
$ cd examples/rest-api-client-demo
$ yarn run-script app getGeneralNotifications
```

## Checklist


packages/rest-api-client/src/client/types/app/notification.ts
```diff
@@ -0,0 +1,9 @@
+export type GeneralNotificationsForResponse = {
+  entity: Record<"type" | "code", string>;
```
shisama, line 2: I think it is good that the values of `type` are literal types, not a string.
You can use the `Entity` type and connect it with the `FIELD_ENTITY` type like the below.
https://github.com/kintone/js-sdk/blob/e0d502e64158fad2a50fd1b6a1a946fbeaa3a2ee/packages/rest-api-client/src/client/types/app/right.ts#L70

```suggestion
  entity:
    | Entity
    | {
        type: "FIELD_ENTITY";
        code: string;
      };
```

NOTE: You cannot apply this suggestion directly because it needs importing the `Entity` type.

shisama (commented):
@chick-p Thank you for your good work!! I left a nit comment. Will you please take a look at it. Thanks :)

chick-p:
@shisama  Thank you reviewing!
I've fixed it.

zaki-yama (approved):
👍

shisama (approved):
💯

PR #606, opened 2021-01-15
feat(rest-api-client): add GET /k/v1/app/notifications/reminder.json

shisama:
## Why

New REST API `GET /k/v1/app/notifications/reminder.json` has already been added.
https://developer.kintone.io/hc/en-us/articles/900003958666

This adds a new method into KintoneRestAPIClient for the new REST API.

## What

This adds `getReminderNotifications()` method into `client.app` object.
The following is a TypeScript sample code for the method.

```typescript
import { KintoneRestAPIClient } from "@kintone/rest-api-client";
const main = async(baseUrl: string, auth: Auth, app: App) => {
  const client = new KintoneRestAPIClient({ baseUrl, auth });
  console.log(await this.client.app. getReminderNotifications({ app }));
}
```
## How to test

This includes tests for this implementation. So just run the below.
`yarn test`

## Checklist


packages/rest-api-client/src/client/types/app/notification.ts
```diff
@@ -33,3 +33,20 @@ export type PerRecordNotificationForResponse = {
   }>;
   revision: string;
 };
+
+export type ReminderNotificationForResponse = {
+  timing: {
+    code: string;
+    daysLater: number | string;
+    hoursLater: number | string;
```
zaki-yama, line 41:
```suggestion
    daysLater: string;
    hoursLater: string;
```

Since this is the type for **response**, I thought the returned values are always string, right?

packages/rest-api-client/src/client/types/app/notification.ts
```diff
@@ -33,3 +33,20 @@ export type PerRecordNotificationForResponse = {
...
+    time: string;
```
zaki-yama, line 42: According to the documentation, I thought `hoursLater` and `time` seemed optional.

zaki-yama (commented):
@shisama 
Thank you!
I confirmed it works as expected.
I left a few comments about response type 👍

shisama:
@zaki-yama Thanks I've fixed them.

zaki-yama (approved):
👍 🚀

PR #609, opened 2021-01-18
feat(rest-api-client): implement GET /k/v1/app/notifications/perRecord.json

zaki-yama:
## Why

In January update of Kintone, several APIs were newly added.
In this PR, I'll support `GET /k/v1/app/notifications/perRecord.json`.
(defined in https://developer.cybozu.io/hc/ja/articles/900004873723)

## What

- [x] implement `getPerRecordNotifications()` method in AppClient
- [x] add a test
- [x] add an example script
- [ ] write a document

## How to test

```
$ yarn build
$ cd examples/rest-api-client-demo
$ yarn run-script app getPerRecordNotifications
```

## Checklist


packages/rest-api-client/docs/app.md
```diff
@@ -1038,6 +1039,36 @@ Gets the [General Notification](https://get.kintone.help/k/en/user/app_settings/
...
+| notifications[].targets[].entity.type | String  | The type of the entity the Per Record Notification settings are configured to.<ul><li>`USER` : User</li><li>`GROUP` : Group</li><li>`ORGANIZATION` : Department</li><li>`FIELD_ENTITY` : Created By, Updated By, User Selection, Group Selection, and Department Selection Fields</li></ul> |
+| notifications[].targets[].entity.code | String  | The code of the entity the Per Record Notification settings are configured to. The following entities will return the following values:<ul><li>`FIELD_ENTITY` : The field code of the field</li><li>For all other entities : The code of the User, Group, or Department</li></ul>           |
+| notifications[].targets[].includeSubs | Boolean | The "Include affiliated departments" setting of the Department. Will always return `false` unless the `notifications[].targets[].entity.type` is set to `ORGANIZATION` or `FIELD_ENTITY` for a Department Selection Field.                                                                  |
```
shisama, line 1065: Why did you abbreviate the below sentences?

```
- true : Affiliated departments do inherit the Per Record Notification settings
- false : Affiliated departments do not inherit the Per Record Notification settings
```

examples/rest-api-client-demo/src/app.ts
```diff
@@ -458,6 +458,28 @@ export class App {
     }
   }
 
+  public async getPerRecordNotifications() {
+    try {
+      const res = await this.client.app.getPerRecordNotifications({
+        app: APP_ID,
+        preview: true,
```
shisama, line 465: Should we create another function when the `preview` is true like other functions?

https://github.com/kintone/js-sdk/blob/29c26551ce4a65d0e800617a8d5852784551a2e8/examples/rest-api-client-demo/src/app.ts#L12-L28

shisama (commented):
Thank you for your good work. I left comments. Please take a look at them.

examples/rest-api-client-demo/src/app.ts
zaki-yama: @shisama 
Oh, I missed that the existing methods have both preview: true and preview: false by independent methods.
You're right. I'll rename this to `~Preview()`, and add another method without preview:true

packages/rest-api-client/docs/app.md
zaki-yama: @shisama 
I thought we'd omitted the description of options if its boolean previously,
(It's probably because the meaning of boolean value is obvious and that sentence seems a bit verbose)

e.g.) `getAppAcl()`'s return properties https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/docs/app.md#returns-15

But it looks `getGeneralNotifications()` and `updatePerRecordNotifications()` both have that sentences, so I'll follow them.

zaki-yama:
@shisama 
Thank you for your review!
I fixed where you commented on.

shisama (approved):
👍 
Maybe you are right to omit the explanation for the value of `includeSubdomains`. We might have to discuss if it can be omitted.

PR #619, opened 2021-01-25
chore: Introduce TypeScript Project References

### LINKED ISSUE #185: TypeScript Project References
https://www.typescriptlang.org/docs/handbook/project-references.html

Currently, we compile TypeScript each package separately, which is not a problem now, but it might be a problem in the future we've had many packages in this repository.

I guess that TypeScript Project References solves the problem., so it's great to investigate it.

koba04:
## Why

Fixes #185 

This enables TypeScript's Project References to build TS files in all packages.
Currently, we have build scripts each package, it's not only inefficient but also problematic with packages have dependencies that are in `js-sdk`.

For example, `customize-uploader` has `rest-api-client` as a dependency. When we have to modify `rest-api-client` for `customize-uploader`, we have to build `rest-api-client` in addition to `customize-uploader`. Otherwise, `rest-api-client` wouldn't be compiled.

With TypeScript Project References, `tsc` does this automatically for us, which means that `yarn start` is the only step to start development. This reduces complexity and makes compiling TS to JS efficient.

After this PR has been merged, we just run `yarn start` to start developing any packages.

## What

I've added TypeScript Project References setting.

I've also updated build scripts. Please check `CONTRIBUTING.md` for more details.

## Checklist


koba04:
This is now ready for review 🚀

chick-p (commented):
@koba04 
Thank you for your great work!
I've left a question. Please check it.

packages/webpack-plugin-kintone-plugin/tsconfig.json
```diff
@@ -1,17 +1,13 @@
...
+    "rootDir": "./src",
     "typeRoots": [
       "node_modules/@types",
       "../../node_modules/@types",
       "types"
-    ], 
+    ]
   },
-  "indlude": ["src/**/*.ts"]
+  "references": []
```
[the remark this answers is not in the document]
koba04: @chick-p Thank you for your feedback. That's a good point 👍  The reason we should ignore `kintone-plugin-packer` is that it isn't written by TypeScript, so we cannot include it in the Project Reference 😭 

I expect that we can include it after #671 has been merged.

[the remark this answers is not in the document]
chick-p: @koba04 
Thank you for teaching me!
I was forgotten what  `kintone-plugin-uploader` is not written  by TypeScript :sweat_drops:

chick-p (approved):
@koba04 
Thank you for answering.
I confirm modified npm scripts and each configuration files (and check script).
This PR looks good to me 👍

examples/rest-api-client-demo/package.json
```diff
@@ -13,6 +13,8 @@
     "directory": "examples/rest-api-client-demo"
   },
   "scripts": {
+    "prebuild": "rimraf dist",
+    "build": "tsc --build",
```
chick-p, line 17:
```suggestion
    "build": "tsc --build --force",
```

Does need `--force` option?

chick-p (approved):
@koba04 
Thank you for supporting to `plugin-packer`, too!

CONTRIBUTING.md
```diff
@@ -50,11 +60,44 @@ If you'd like to release a new package.
...
+In most cases, the script would be `yarn clean`, which runs `rimraf` for build assets directories.
+
+`postbuild` is run after `build`, so it's useful to build other assets like ESM and UMD builds.
+
+### `lint`
+
+You have to define `lint` script to lint source files.
+In most cases, the script would be the `eslint` command.
+
+You don't have to `tsc --noEmit` as the `lint` script because `tsc` is run as the `build` script that is run before the `lint` script.
```
shisama, line 83: @koba04 Sorry for my comment after merging this. You removed `lint:ts` command for `tsc --noEmit`. So, Is it OK that CI won't run the type check that does previously?

koba04: @shisama That's a good point.
Yes, the type check is run as `yarn build` and CI runs `yarn build` before `yarn lint`.
https://github.com/kintone/js-sdk/blob/6f5f2895757e70bb063b4719601f9e1a43e61b61/.github/workflows/lint.yml#L24

If you run `yarn lint` on your local machine without running `yarn build`, you would see `node/no-missing-require` errors because the rule checks if a module is there. Of course, we could run `yarn build` as the `prelint` and `pretest` scripts, but it is often unnecessary and makes script execution time longer.
I expect that we always  run `yarn start` as the development workflow on `js-sdk`. In that case, we don't have to run `yarn build` as `prelint` and `pretest`.

shisama: @koba04 I see. I missed `yarn build` in lint.yml. I agree that `prelint` and `pretest` are not necessary. Thanks :)

PR #623, opened 2021-01-26
chore: fix the way to use cybozu/renovate-config

koba04:
## Why

We use `cybozu/renovate-config` as `@cybozu`, but it's not an official way that the document describes.
https://docs.renovatebot.com/config-presets/

So I've updated the way.

## Checklist


shisama (approved):
Thanks!

PR #654, opened 2021-02-02
docs(data-loader): options in data-loader's readme

shisama:
## Why

The default value for the `--password` option indicated in README is incorrect.
Also, the `--api-token` option is not listed in README even though it exists as an option.

## What

The default value for the `--password` option in README was fixed.
The `--api-token` option was listed.

## How to test

```
$ yarn lint
```

## Checklist

### DIFF
--- a/packages/data-loader/README.md
+++ b/packages/data-loader/README.md
@@ -36,7 +36,8 @@ import \
 Options:
   --base-url             Kintone Base Url           [default: process.env.KINTONE_BASE_URL]
   --username, -u         Kintone Username           [default: process.env.KINTONE_USERNAME]
-  --password, -p         Kintone Password           [default: process.env.KINTONE_USERNAME]
+  --password, -p         Kintone Password           [default: process.env.KINTONE_PASSWORD]
+  --api-token            App's API token
   --basic-auth-username  Kintone Basic Auth Username
   --basic-auth-password  Kintone Basic Auth Password
   --app                  The ID of the app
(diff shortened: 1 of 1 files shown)


chick-p:
@shisama 
Thank you for fixing README!
It look like good 👍 

Incidentally, some options support environment values as default.
Ref. https://github.com/kintone/js-sdk/blob/master/packages/data-loader/src/commands/import.ts
- `--api-token`
- `--basic-auth-username`
- `--basic-auth-password`
- --`guest-space-id`

Would you like to write these?

b4h0-c4t (commented):
It seems unnatural that only --attachment-dir is restricted by "type".
How about removing the type of yargs option in `commands/export.ts`?

shisama:
@chick-p Thanks! They have been added.

shisama:
@b4h0-c4t 
> How about removing the type of yargs option in commands/export.ts?

Thank you for your comment.
The `type` of yargs option was added in https://github.com/kintone/js-sdk/commit/aae68dd79bfd991df60c4019092120f19b988eee.
`--attachment-dir` option has `type` to validate if it is a string or not. When we made the `--attachment-dir` optional, it comes to be necessary to check it. If not validating it, it can accept a boolean value when `--attachment-dir` has no value like: `kintone-data-loader export --app 1 --attachment-dir`.
That's why I think the `type` option in `command/export.ts` is necessary.
Will you please tell me another idea if you have it? Thanks :)

packages/data-loader/README.md
```diff
@@ -64,15 +65,17 @@ export \
...
-  --base-url             Kintone Base Url           [default: process.env.KINTONE_BASE_URL]
-  --username, -u         Kintone Username           [default: process.env.KINTONE_USERNAME]
-  --password, -p         Kintone Password           [default: process.env.KINTONE_USERNAME]
-  --basic-auth-username  Kintone Basic Auth Username
-  --basic-auth-password  Kintone Basic Auth Password
+  --base-url             Kintone Base Url            [default: process.env.KINTONE_BASE_URL]
+  --username, -u         Kintone Username            [default: process.env.KINTONE_USERNAME]
+  --password, -p         Kintone Password            [default: process.env.KINTONE_PASSWORD]
+  --api-token            App's API token
+  --api-token            App's API token             [default: process.env.KINTONE_API_TOKEN]
```
chick-p, line 72:
```suggestion
  --api-token            App's API token             [default: process.env.KINTONE_API_TOKEN]
```

chick-p (commented):
@shisama 
Thank you!
I left a comment, please check it.

b4h0-c4t:
@shisama 
> Will you please tell me another idea if you have it? Thanks :)

It felt a little uneven because there was no type check on the other values.
If you plan to check type on the other options, would be willing to implement that in another PR.

But, it's not a very important problem, so I think it's fine to move forward as is.

b4h0-c4t (approved):

chick-p (approved):
@shisama 
👍

shisama:
@b4h0-c4t Yep, I will create another PR to add types to other options. Thanks.

PR #679, opened 2021-02-10
chore(data-loader): download attachments sequentially

zaki-yama:
## Why

For performance reason, we want to avoid downloading attachments concurrently.

## How to test

- Create a Kintone app, insert records, and upload attachments to some of the records
- Run the following commands:

```zsh
$ cd packages/data-loader
$ yarn build
$ node cli.js export --app=<APP_ID> --attachment-dir=foo # and username, password, base-url
```

- Then check the attachments are downloaded

## Checklist

### DIFF
--- a/packages/data-loader/src/controllers/export.ts
+++ b/packages/data-loader/src/controllers/export.ts
@@ -1,6 +1,5 @@
 import { promises as fs } from "fs";
 import path from "path";
-import PQueue from "p-queue";
 
 import { KintoneRestAPIClient } from "@kintone/rest-api-client";
 import { AppID, Record } from "@kintone/rest-api-client/lib/client/types";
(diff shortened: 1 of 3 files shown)


packages/data-loader/src/controllers/export.ts
```diff
@@ -77,14 +62,19 @@ const getFileInfos = (record: Record) => {
...
 ) => {
-  const { fileKey, name } = fileInfo;
-  const file = await apiClient.file.downloadFile({ fileKey });
-  const recordId = record.$id.value as string;
-  const dir = path.resolve(attachmentDir, recordId);
-  await fs.mkdir(dir, { recursive: true });
-  await fs.writeFile(path.resolve(dir, name), Buffer.from(file));
+  for (const record of records) {
+    const fileInfos = getFileInfos(record);
+    for (const fileInfo of fileInfos) {
```
zaki-yama, line 70: I moved 2 loops (1 for `records`, 1 for `fileInfos`) to `downloadAttachments` because

- want to simplify the logic of `exportRecords`
- original `downloadAttachments()` actually downloads 1 file, so it's not `downloadAttachment"s"`

PR #719, opened 2021-02-25
feat(data-loader): Support CSV format

nakajmg:
## Checklist


packages/data-loader/src/controllers/import.ts
```diff
@@ -39,10 +42,23 @@ export const buildImporter = ({
     const { app, filePath } = options;
     const buf = await fs.readFile(filePath);
     const type = extractFileType(filePath);
-    const records = parser(type, buf.toString());
+    const parser = buildParser(type);
+
+    const records =
+      type === "csv"
```
shisama, line 48: It may be better to move this boolean expression to the parser phase. But I'm not sure if it should be in the parser phase...

packages/data-loader/src/printers/csvPrinter.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const extractFieldCodes = (records: KintoneRecords) => {
+  const firstRecord = records.slice().shift();
+  if (!firstRecord) return [];
+  return Object.keys(firstRecord).filter((key) =>
+    isSupportedFieldType(firstRecord[key])
+  );
+};
+
+const lexer = (field: KintoneRecordField.OneOf) => {
```
zaki-yama, line 42: Personally, the term `lexer` seems it accepts string and generate tokens.
But what it does is to accept kintone's field (a structured object, not string) and extract its value.

How about renaming it?
I came up with: `parseField(field)`, `extractValue(field)`, `valueOf(field)` 

(If you refer to something, please tell me!)

packages/data-loader/src/printers/csvPrinter.ts
```diff
@@ -0,0 +1,91 @@
...
+    case "CALC":
+    case "UPDATED_TIME":
+    case "CREATED_TIME":
+      return encloseInQuotation(field.value);
+    case "CREATOR":
+    case "MODIFIER":
+      return encloseInQuotation(field.value.code);
+    case "MULTI_SELECT":
+    case "CHECK_BOX":
+      return encloseInQuotation(field.value.join(LINE_BREAK));
```
zaki-yama, line 61: Just a question: Are there any field types which we shouldn't apply `encloseInQuatation()` ?
If not, we can apply it outside of this function.

packages/data-loader/src/printers/csvPrinter.ts
```diff
@@ -0,0 +1,91 @@
...
+export const convertKintoneRecordsToCsv = (records: KintoneRecords) => {
+  const fieldCodes = extractFieldCodes(records);
+
+  const header = fieldCodes
+    .map((fieldCode) => encloseInQuotation(fieldCode))
+    .join(SEPARATOR);
+
+  const rows = records
+    .slice()
+    .reverse()
```
zaki-yama, line 76: I couldn't understand why you apply `slice()` and `reverse()` here.
Since the order of records should be determined by the user (we need to accept `order by` clause in the future), I thought we shouldn't change their order, right?

packages/data-loader/src/printers/csvPrinter.ts
```diff
@@ -0,0 +1,91 @@
...
+    .map((record) => {
+      return fieldCodes
+        .map((fieldCode) => lexer(record[fieldCode]))
+        .join(SEPARATOR);
+    });
+
+  return (
+    [header, ...rows].join(LINE_BREAK).replace(/((?!.)\s)+/g, LINE_BREAK) +
```
zaki-yama, line 84:
```
.replace(/((?!.)\s)+/g, LINE_BREAK)
```

I'd be happy if there's a comment to describe how it does, and why it's necessary.

packages/data-loader/src/printers/__tests__/csvPrinter.test.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const expectedCsv = `"recordNumber","updatedTime","dropDown","creator","modifier","richText","singleLineText","number","radioButton","multiLineText","createdTime","checkBox","calc","multiSelect"
+"9","2021-02-16T02:43:00Z","sample1","username","username","<div><div>rich text editor<br /></div></div><div>rich text editor<br /></div><div>rich text editor<br /></div>","""single line text""","8","sample2","multi
+line
+text","2021-02-10T06:14:00Z","""sample2""","16","""sample3""
+sample4"
+`;
+
+describe("csvPrinter", () => {
+  it("should convert kintone records to csv string correctly", async () => {
```
zaki-yama, line 16: is `async` necessary?

packages/data-loader/src/printers/csvPrinter.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+const escapeQuotation = (fieldValue: string) => fieldValue.replace(/"/g, '""');
+
+const encloseInQuotation = (fieldValue: string | null) =>
+  `"${fieldValue ? escapeQuotation(fieldValue) : ""}"`;
+
+const extractFieldCodes = (records: KintoneRecords) => {
+  const firstRecord = records.slice().shift();
+  if (!firstRecord) return [];
```
zaki-yama, line 36: Do you have any reasons to create a shallow copy of all records here?
I thought we can use just `records[0]` since we use only the first record, and the function doesn't modify the record.

packages/data-loader/src/printers/csvPrinter.ts
```diff
@@ -0,0 +1,91 @@
...
+    "MODIFIER",
+    "UPDATED_TIME",
+    "CREATED_TIME",
+    "MULTI_SELECT",
+    "CHECK_BOX",
+  ];
+  return supportedFieldTypes.includes(field.type);
+};
+
+const escapeQuotation = (fieldValue: string) => fieldValue.replace(/"/g, '""');
```
zaki-yama, line 29: nits
I just wondered which term is more suitable, `quatation` or `quote(s)`.

According to MDN, `quote` is better?
https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Strings#single_quotes_vs._double_quotes

`escapeDoubleQuotes()` might be better because it's clear that the function doesn't affect to single quotes.

shisama (commented):
@nakajmg Thank you for your excellent work. I left some comments. However, some of them are just comments and ideas so that they are not necessary to be fixed.

packages/data-loader/src/printers/csvPrinter.ts
nakajmg: > But what it does is to accept kintone's field (a structured object, not string) and extract its value.

It will be renamed to `extractFieldValue` in the next version 💁‍♀️

https://github.com/kintone/js-sdk/pull/760/files#diff-9e23ea4c64b01417857eb0d4828b75fca98d70dccf53507f5aaf19344da5725bR6

packages/data-loader/src/printers/csvPrinter.ts
nakajmg: This version can be applied on the outside.
However in order to support SUBTABLE, should handle it inside.

https://github.com/kintone/js-sdk/pull/760/files#diff-9e23ea4c64b01417857eb0d4828b75fca98d70dccf53507f5aaf19344da5725bR28-R42

packages/data-loader/src/printers/csvPrinter.ts
nakajmg: > `slice()`

I wanted to avoid mutation.

> we need to accept order by clause in the future

I didn't consider that. I'll fix it.

packages/data-loader/src/printers/csvPrinter.ts
nakajmg: This is to unify line feed codes. (`\n`)

Is this easier to understand?

```js
[header, ...rows].join(LINE_BREAK).replace(/\r?\n/g, LINE_BREAK) +
```

or extract to function?

packages/data-loader/src/printers/csvPrinter.ts
nakajmg: > any reasons to create a shallow copy

I don't like access by index directly like a magic number.
However it wasteful.

I'll change it.

FYI: In next version, header fields is created by `fields.json` instead of record.
https://github.com/kintone/js-sdk/pull/760/files#diff-62a400cffc91b23a9bd6eb112044e90b7532c06ee5137fed3d3da62acb810f40R6-R28

packages/data-loader/src/printers/csvPrinter.ts
nakajmg: > escapeDoubleQuotes() might be better because it's clear that the function doesn't affect to single quotes.

👍

packages/data-loader/src/printers/csvPrinter.ts
zaki-yama: I understand. Thanks!

shisama:
@nakajmg Thanks! Code looks good to me. I think this will be ready to merge after adding documents.

packages/data-loader/src/controllers/import.ts
```diff
@@ -39,10 +42,23 @@ export const buildImporter = ({
...
+        ? parser(buf.toString(), await fetchFieldsJson(options))
+        : // @ts-ignore
+          parser(buf.toString());
     await addAllRecordsChunk(app, records);
   }
 
+  async function fetchFieldsJson(options: Options) {
+    const fieldsJson: {
+      properties: Record<string, KintoneFormFieldProperty.OneOf>;
+    } = await apiClient.app.getFormFields(options);
```
zaki-yama, line 58: `getFormFields` can accept generics like this: 

```suggestion
    const fieldsJson = await apiClient.app.getFormFields<
      Record<string, KintoneFormFieldProperty.OneOf>
    >(options);
```

The result of the type `fieldsJson.properties` is the same, but I think it's more natural way to specify its type.

PR #726, opened 2021-03-01
fix(rest-api-client): prevent infinite loop when `condition` contains OR clause

### LINKED ISSUE #716: Continuously send requests when use getAllRecordsWithId
## Target Package

- @kintone/rest-api-client 1.10.0

## Environment

- Operating System: Windows 10
- Browser: Chrome 88.0.4324.182
- Node.js version: v14.15.3

## Current Behavior

Step 1. Create a new kintone app, no field is OK.
Step 2. Add more than 500 records.
Step 3. On JS Customization page of this APP, add the UMD file of rest-api-client.
Step 4. On list page of this APP, open DevTools of browser and run following code.
`const client = new KintoneRestAPIClient();
client.record
  .getAllRecordsWithId({ app: kintone.app.getId(), condition: '$id != "" or $id = ""' })
  .then((resp) => {
    console.log(resp);
  })
  .catch((err) => {
    console.log(err);
  });`

=> Bug occurs:
The API requests are send continuously.
Always return the first 500 records.

**Memo:**
If condition is '($id != "" or $id = "")', this bug does not occur.

I think adding () in rest-api-client is better.
https://github.com/kintone/js-sdk/blob/f54a4ccbaec43f90dceaf6533431f4951d37fa92/packages/rest-api-client/src/client/RecordClient.ts#L274
const conditionQuery = condition ? \`${condition} and \` : "";
->
const conditionQuery = condition ? \`**(**${condition}**)** and \` : "";

## Expected Behavior

Do not send requests continuously.
Return the correct result.

zaki-yama:
## Why

Fixes #716 .

## What

Enclose `condition` parameter of `getAllRecords()` and `getAllRecordsWithId()` with parenthesis (`()`). (See e634fb4)
NOTE: I didn't enclose the `conditionit` parameter of `getAllRecordsWithOffset()`.

## How to test

1. Create a kintone app, insert more than 500 records
2. Call `getAllRecords()` with `condition` parameter that meets the following conditions:
    - `condition` contains `or` clause (e.g. `'Customer != "foo" or Status in ("Completed")'` )
    -  the number of filtered records also exceeds 500 records
3. Confirm that all expected records are returned, and the process finishes correctly

You can test it easily using [`getAllRecords()` demo script](https://github.com/kintone/js-sdk/blob/%40kintone/rest-api-client%401.10.0/examples/rest-api-client-demo/src/record.ts#L161) with modified `condition` parameter.

## Checklist

- [x] Added tests if it is required. => See 0c810bc

### DIFF
--- a/packages/rest-api-client/src/client/RecordClient.ts
+++ b/packages/rest-api-client/src/client/RecordClient.ts
@@ -271,7 +271,7 @@ export class RecordClient {
     const GET_RECORDS_LIMIT = 500;
 
     const { condition, ...rest } = params;
-    const conditionQuery = condition ? `${condition} and ` : "";
+    const conditionQuery = condition ? `(${condition}) and ` : "";
     const query = `${conditionQuery}$id > ${id} order by $id asc limit ${GET_RECORDS_LIMIT}`;
     const result = await this.getRecords<T>({ ...rest, query });
     const allRecords = records.concat(result.records);
--- a/packages/rest-api-client/src/client/__tests__/RecordClient.test.ts
+++ b/packages/rest-api-client/src/client/__tests__/RecordClient.test.ts
@@ -481,9 +481,9 @@ describe("RecordClient", () => {
           params: {
             app: params.app,
             fields: params.fields,
-            query: `${
+            query: `(${
               params.condition || ""
-            } and $id > 0 order by $id asc limit 500`,
+            }) and $id > 0 order by $id asc limit 500`,
           },
         });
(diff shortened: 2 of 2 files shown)


zaki-yama:
I refactored a little in RecordClient.test (1f765e2).
The actual change is 0c810bc.

chick-p (approved):
@zaki-yama 
Thank you for fixing!

shisama (approved):
@zaki-yama Thank you for your work. Code looks good to me. However, does this modification occur a breaking change because the result will change by changing the condition?

PR #822, opened 2021-04-20
feat(rest-api-client): add an option to set a custom User-Agent string in Node.js environment

chick-p:
## Why
In the Node.js environment, a User-Agent has been set a fixed value as `Node.js/v12.21.0(Darwin) @kintone/rest-api-client@1.12.3`.
However, I want to set any value of User-Agent.

According to https://developer.kintone.io/hc/en-us/articles/212494698#consideration, it announces that set an appropriate value in the User-Agent header.

> Set an appropriate user agent
> Please set an appropriate value in the User-Agent header, so that we can identify what services or tools the requests are coming from.

## What

set an appropriate value in the User-Agent header as `KintoneRestApiClient` option.

## How to test

yarn test

## Checklist


koba04 (approved):
Great work, LGTM 👍 👍 👍

PR #832, opened 2021-04-23
fix(data-loader): apply accurated types to CLI command builder

mshrtsr:
## Why

In this PR, I apply accurated types to CLI command builder (`builder` function in `src/commands/*.ts`)
This resolves following problems.
- type inference does not work in command builder function because its parameter is `any`.
- type inference is breaking between `src/main.ts` and `src/commands/*.ts`bacause of using of `yargs.commandDir`.

## What

- [x] remove `any` from CLI command builder
- [x] use `yargs.command` instead of `yargs.commandDir`
- [x] infer the parameter type of handler from return type of builder

## How to test

This PR doesn't affect on runtime.

```sh
$ yarn build
$ yarn lint
$ yarn test
```

## Checklist

### DIFF
--- a/packages/data-loader/src/commands/export.ts
+++ b/packages/data-loader/src/commands/export.ts
@@ -83,7 +83,7 @@ export const builder = (args: yargs.Argv) =>
     });
 
 type Args = yargs.Arguments<
-  ReturnType<typeof builder> extends yargs.Argv<infer U> ? U : unknown
+  ReturnType<typeof builder> extends yargs.Argv<infer U> ? U : never
 >;
 
 export const handler = (args: Args) => {
--- a/packages/data-loader/src/commands/export.ts
+++ b/packages/data-loader/src/commands/export.ts
@@ -14,7 +14,6 @@ export const builder = (args: yargs.Argv) =>
       default: process.env.KINTONE_BASE_URL,
       defaultDescription: "KINTONE_BASE_URL",
       type: "string",
-      demandOption: true,
     })
     .option("username", {
       alias: "u",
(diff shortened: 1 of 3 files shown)


packages/data-loader/src/commands/import.ts
```diff
@@ -71,6 +66,26 @@ export const builder = (yargs: any) =>
     .option("pfx-file-password", {
       describe: "The password of client certificate file",
       type: "string",
-    });
+    })
+    // NOTE: Unexpected `undefined` in inferred type by `option()`.
```
zaki-yama, line 70: It'd be more helpful if we can understand that it's just a workaround, and incorrect type inferrence was caused by yargs.
So... what about something like this?

```suggestion
    // NOTE: Since yargs doesn't detect the type correctly by adding `demandOption: true` in `option()`,
    // (inferred type always contains `| undefined`)
    // we declare the required params later as a workaround.
```

packages/data-loader/src/commands/export.ts
```diff
@@ -82,6 +78,28 @@ export const builder = (yargs: any) =>
     .option("pfx-file-password", {
       describe: "The password of client certificate file",
       type: "string",
-    });
+    })
+    // NOTE: Unexpected `undefined` in inferred type by `option()`.
```
zaki-yama, line 82: see: https://github.com/kintone/js-sdk/pull/832/files#r621817312 🙏

mshrtsr:
@zaki-yama 
Thank you for the review!
Your suggestion changes are looks good, so I commited it in 42eaaa6.
I also added related issue yargs/yargs#1928 to the NOTE comment.

zaki-yama (approved):
👍

PR #843, opened 2021-04-28
chore(data-loader): set a User-Agent string for data-loader

chick-p:
## Why

According to https://developer.kintone.io/hc/en-us/articles/212494698#consideration, it announces that set an appropriate value in the User-Agent header.
I want that data-loader sets a custom User-Agent string.

## What

set a custom User-Agent string as `data-loader@version`

## How to test

`yarn test`

## Checklist


zaki-yama (approved):
👍 👍 👍

PR #854, opened 2021-04-30
docs(rest-api-client): add "or `null`" for nullable properties in Type column

### LINKED ISSUE #827: Description of return value types in the doc
## Target Doc

- [js-sdk/packages/rest-api-client/docs/record.md](https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/docs/record.md)

## Current Behavior

- The return type of `totalCount` is described as `String` even though the value of `totalCount` will be `null` when the request parameter `totalCount` is set to `false` or ignored. I assume that the description should be `String` or `null`.

  - getRecords - Returns
    https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/docs/record.md#returns-4

- The return type of the `totalCount` is `String` even though the type is described as `Integer` on the reference page. I did not confirm which one is true, but we should fix either of them.

  - createCursor - Returns
    https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/docs/record.md#returns-18

  - createCursor - Reference
    https://github.com/kintone/js-sdk/blob/master/packages/rest-api-client/docs/record.md#reference-10

zaki-yama:
## Why

Fixes #827 .
It's more convenient if we can understand nullable properties by looking at Type column.

## What
Add "or `null` to Type column to the document.

## Checklist


zaki-yama:
For reviewers:
Please see the diff by switching to the rich diff view.
SS 2021-04-30 15 57 31

nakajmg (approved):

PR #887, opened 2021-05-18
feat(data-loader): generate attachment metadata when export (JSON)

mshrtsr:
## Why

Refer to #950.

This PR resolves two tasks of #950.
- Design the format of the value of attachment file in JSON
- Consider duplicate name of attachments in the same field

## What

- [x] generate `attachments.json` when export
- [x] fix existing tests
- [x] add tests

### Metadata spec

- Attachment metadata will be written to `<attachment-dir>/attachments.json`
- Attachment schema is written in [attachment.ts](https://github.com/kintone/js-sdk/pull/887/files#diff-827e45a0ece2af288fafa1cb00e89dd8603b38cdbebecfed2f77cd4c49ea1af9)
- Duplicate file names in a record will be renamed with the index as `<original filename>-<index>.<extension>`

#### Schema example

<details>

If the following records are exported,

```json
[
  {
    "Attachment_0": {
      "type": "FILE",
      "value": [
        {
          "fileKey": "...",
          "name": "image.png",
          "contentType": "image/png",
          "size": "..."
        },
        {
          "fileKey": "...",
          "name": "image.png",
          "contentType": "image/png",
          "size": "..."
        }
      ]
    },
    "Table_0": {
      "type": "SUBTABLE",
      "value": [
        {
          "id": "...",
          "value": {
            "添付ファイル": {
              "type": "FILE",
              "value": [
                {
                  "fileKey": "...",
                  "name": "image.png",
                  "contentType": "image/png",
                  "size": "..."
                },
                {
                  "fileKey": "...",
                  "name": "image.png",
                  "contentType": "image/png",
                  "size": "..."
                }
              ]
            }
          }
        }
      ]
    },
  }
]

```

`attachments.json` will be like this.

```json
[
  {
    "Attachment_0": [
      "image.png",
      "image-1.png"
    ],
    "Table_0": [
      {
        "添付ファイル": [
          "image-2.png",
          "image-3.png"
        ]
      }
    ]
  }
]
```

</details>

## How to test

```sh
$ yarn build
$ yarn test
```

```sh
$ cd packages/data-loader
$ yarn build
$ node cli.js export --app=<APP_ID> --attachment-dir=foo # and username, password, base-url
```

## Checklist

### DIFF
--- a/packages/data-loader/src/controllers/__tests__/export.test.ts
+++ b/packages/data-loader/src/controllers/__tests__/export.test.ts
@@ -140,7 +140,7 @@ describe("export", () => {
         type: "SINGLE_LINE_TEXT",
         value: "value1",
       },
-      subTable: {
+      subtable: {
         type: "SUBTABLE",
         value: [
           {
@@ -182,7 +182,7 @@ describe("export", () => {
 
     const attachmentMetadataList = [
       {
-        subTable: [{ attachmentInSubtable: ["test.txt"] }],
+        subtable: [{ attachmentInSubtable: ["test.txt"] }],
       },
       {},
     ];
@@ -208,7 +208,7 @@ describe("export", () => {
       path.join(
         tempDir,
         recordWithAttachment.$id.value,
-        recordWithAttachment.subTable.value[0].value.attachmentInSubtable
+        recordWithAttachment.subtable.value[0].value.attachmentInSubtable
           .value[0].name
       )
     );
@@ -253,7 +253,7 @@ describe("export", () => {
           },
         ],
       },
-      subTable: {
+      subtable: {
         type: "SUBTABLE",
         value: [
           {
@@ -317,7 +317,7 @@ describe("export", () => {
       {
         attachment: ["test.txt", "test-1.txt"],
         attachment2: ["test-2.txt", "test-3.txt"],
-        subTable: [
+        subtable: [
           {
             attachmentInSubtable: ["test-4.txt", "test-5.txt"],
             attachmentInSubtable2: ["test-6.txt", "test-7.txt"],
--- a/packages/data-loader/src/controllers/__tests__/export.test.ts
+++ b/packages/data-loader/src/controllers/__tests__/export.test.ts
@@ -214,6 +214,137 @@ describe("export", () => {
     );
     expect(downloadedFile.toString()).toBe(testFileData);
   });
+  it("can download files with duplicate file names to a specified directory", async () => {
+    const recordWithAttachment = {
+      $id: {
+        value: "2",
+      },
+      fieldCode: {
+        type: "SINGLE_LINE_TEXT",
+        value: "value1",
+      },
+      attachment: {
+        type: "FILE",
+        value: [
+          {
+            contentType: "text/plain",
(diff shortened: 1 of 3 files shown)


packages/data-loader/src/controllers/__tests__/export.test.ts
```diff
@@ -95,6 +95,13 @@ describe("export", () => {
       },
     ];
 
+    const attachmentMetaDataList = [
```
zaki-yama, line 98: MetaData -> Metadata

There might be more in other places.

packages/data-loader/src/controllers/attachments.ts
```diff
@@ -0,0 +1,15 @@
...
+
+type Record = {
+  [fieldCode: string]: Field;
+};
+
+type Field = FileField | SubTableField;
+
+type FileField = string[];
+
+export type SubTableField = SubTableRow[];
```
zaki-yama, line 11: SubTable -> Subtable

There might be more in other places.

packages/data-loader/src/controllers/__tests__/export.test.ts
mshrtsr: fixed in 338c939

packages/data-loader/src/controllers/attachments.ts
mshrtsr: fixed in 4897922

mshrtsr:
@zaki-yama 
Thank you for the review!
I commited some fixes about it.

packages/data-loader/src/controllers/export.ts
```diff
@@ -55,52 +56,146 @@ export async function exportRecords(
...
+        return acc;
+      }
+      if (field.type === "SUBTABLE") {
+        acc[fieldCode] = generateAttachmentMetadataInSubtable(
+          field,
+          localFileNameSet
+        );
+        return acc;
+      }
+      return acc;
```
zaki-yama, line 76: Since it always return `acc`, and each `if` statement is short enough to understand,
I thought it'd be easier to understand to quite early return like this:

```suggestion
      if (field.type === "FILE") {
        acc[fieldCode] = field.value.map((fileInformation) =>
          generateUniqueLocalFileName(fileInformation.name, localFileNameSet)
        );
      } else if (field.type === "SUBTABLE") {
        acc[fieldCode] = generateAttachmentMetadataInSubtable(
          field,
          localFileNameSet
        );
      }
      return acc;
```

(I also replaced the second `if` statement with `else  if`)
What do you think?

---

By the way, if I were you I would use `for...of` instead of `reduce` in this case, but it's a matter of taste 👍 

```typescript
const metadata: AttachmentMetaData = {};
for (const [fieldCode, field] of Object.entries(record)) {
  if (field.type === "FILE") {
    metadata[fieldCode] = field.value.map((fileInfo) =>
      generateUniqueLocalFileName(fileInfo.name, localFileNameSet)
    );
  } else if (field.type === "SUBTABLE") {
    metadata[fieldCode] = generateAttachmentMetaDataInSubtable(
      field,
      localFileNameSet
    );
  }
}
return metadata;
```

zaki-yama (approved):

PR #925, opened 2021-06-01
fix(customize-uploader): upload customize files sequentially

chick-p:
## Why

For impact to performance, we want to avoid uploading customize files in parallel.

## What

upload customize files sequentially.

## How to test

```shell
yarn build
bin/cli.js sample/customize-manifest.json
```

## Checklist

### DIFF
--- a/packages/customize-uploader/src/commands/index.ts
+++ b/packages/customize-uploader/src/commands/index.ts
@@ -51,33 +51,32 @@ export async function upload(
     if (!updateBody) {
       console.log(m("M_StartUploading"));
       try {
-        const [desktopJS, desktopCSS, mobileJS, mobileCSS] = await Promise.all(
-          [
-            manifest.desktop.js,
-            manifest.desktop.css,
-            manifest.mobile.js,
-            manifest.mobile.css,
-          ].map((files) =>
-            Promise.all(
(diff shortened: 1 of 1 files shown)


packages/customize-uploader/src/commands/index.ts
```diff
@@ -51,33 +51,32 @@ export async function upload(
...
-        );
+        const uploadFilesResult = [];
+        for await (const files of [
+          manifest.desktop.js,
+          manifest.desktop.css,
+          manifest.mobile.js,
+          manifest.mobile.css,
+        ]) {
+          const results = [];
+          for await (const file of files) {
```
zaki-yama, line 62: Just a question: Are these `await` in 2 for loops needed?
Didn't it run as expected without them?

zaki-yama (commented):
👍 
Thanks! I left a comment.

packages/customize-uploader/src/commands/index.ts
chick-p: @zaki-yama 
I confirmed that It runs without `await`....
I've removed `await`.

chick-p:
@zaki-yama 
Thank you for your review!
I've fixed according to your comment.

PR #1019, opened 2021-07-20
style(dts-gen): remove prettier config

mshrtsr:
## Why

dts-gen has own `prettier.config.js`, but we want to use same config for all js-sdk packages.

## What

- [x] remove `prettier.config.js` from dts-gen package
- [x] fix `lint:fix` command of dts-gen package
- [x] reformat code

In this PR, there is no change in the generated type definition file.

## How to test

```sh
$ yarn lint & yarn test
```

check the generated type definition file.
```sh
$ cd ./packages/dts-gen
$ yarn demo
$ cat ./demo-fields.d.ts
```

## Checklist


chick-p (approved):
💯

Issue #447, opened 2020-07-14
Calc and File field requires type


the-red:
In kintone.d.ts, The `type` property is optional for most field types.
https://github.com/kintone/dts-gen/blob/353ac959a38844f23e4f8fc0f928c9d0263fbd42/kintone.d.ts#L261

However, only `Calc` and `File` are requires `type`. 
https://github.com/kintone/dts-gen/blob/353ac959a38844f23e4f8fc0f928c9d0263fbd42/kintone.d.ts#L289
https://github.com/kintone/dts-gen/blob/353ac959a38844f23e4f8fc0f928c9d0263fbd42/kintone.d.ts#L372

Is there a reason for this?

koba04:
We've merged `dts-gen` into `js-sdk`, so this issue has been transferred.

the-red:
@koba04 
The issue has been auto-closed, will you be able to handle it?

This one was auto-closed as well.
https://github.com/kintone/js-sdk/issues/448

koba04:
@the-red I've re-opened this! You can avoid that the bot closes an issue by putting a comment into the issue.

the-red:
not stale

zaki-yama:
@the-red 
Thank you! I reopened.

the-red:
Not stale

the-red:
Not stale... 😭

shintaroNagata:
@the-red 
We have reopened this issue & #448 .
However, we have not decided on the plan nor the schedule for these issues yet.
Issues that have an "open" state may look to be under development. We think that this situation is undesirable.
To avoid occurring perception gaps between maintainer and contributor,  we will decide and announce the policy for states of issues in the future.
Until we decide, we will keep these issues "open" . Please note that they can be "close" even though they are under consideration depends on the policy we will decide.

PR #1152, opened 2021-10-18
chore: stop using Stale bot

shintaroNagata:
## Why

We've introduced the stale bot to sort out old tickets in this repository since https://github.com/kintone/js-sdk/pull/740 .
However, the bot closes the tickets that we don't want to close automatically.
To make it work as we intended, we think that we should decide the policy for dealing with issues clearly. ref: https://github.com/kintone/js-sdk/issues/447#issuecomment-937451838

## What

delete the config file for stale bot.

## Checklist


mshrtsr (approved):
LGTM 💯

Issue #1231, opened 2021-12-01
Apply ESLint and Prettier to generated *.d.ts file


mshrtsr:
## Package

- @kintone/dts-gen

## Why

relate to kintone/js-sdk-ja#22 .

## What

When dts-gen generates type definition file, apply ESLint & Prettier to generated *.d.ts file.

PR #1266, opened 2021-12-21
feat(dts-gen): apply eslint to generated `*.d.ts file`

chick-p:
## Why

Fixed https://github.com/kintone/js-sdk/issues/1231

## What

apply ESLint to generated `*.d.ts file`

## How to test

```
$ yarn install && yarn build
$ cd packages/dts-gen
$ node dist/index.js --app-id 1 --type-name SampleFields --namespace company.name.types -o sample-fields.d.ts
```

```
$ yarn install && yarn build && yarn test
```

## Checklist


packages/dts-gen/src/templates/template.ts
```diff
@@ -11,9 +12,38 @@ interface RenderInput {
   fieldTypeGroups: FieldTypeGroups;
 }
 
-const renderAsFile = (output: string, renderInput: RenderInput) => {
+const renderAsFile = async (output: string, renderInput: RenderInput) => {
```
ueokande, line 15: replace `fs.writeFileSync` at the bottom of the function with promise-based api.

packages/dts-gen/src/templates/template.ts
```diff
@@ -11,9 +12,38 @@ interface RenderInput {
...
+      if ("output" in r) {
+        return r.output;
+      }
+      if ("source" in r) {
+        return r.source;
+      }
+      return "";
+    })
+    .join("");
+  const prettySource = prettier.format(eslintOutput, {
```
ueokande, line 46: why not use `@cybozu/eslint-config/presets/typescript-prettier`?

packages/dts-gen/src/templates/template.ts
```diff
@@ -11,9 +12,38 @@ interface RenderInput {
...
+      // https://eslint.org/docs/developer-guide/nodejs-api#-lintresult-type
+      if ("output" in r) {
+        return r.output;
+      }
+      if ("source" in r) {
+        return r.source;
+      }
+      return "";
+    })
+    .join("");
```
ueokande, line 45: This joining has a strangeness because we do not want to join the result code here. We expect that the result has a single element, and the method returns it. Access the first element explicitly (check the result count if needed before, but not required).

packages/dts-gen/src/templates/template.ts
```diff
@@ -11,9 +12,38 @@ interface RenderInput {
...
+      },
+      rules: {
+        "@typescript-eslint/no-namespace": [
+          "error",
+          { allowDeclarations: true },
+        ],
+      },
+    },
+  });
+  const eslintResult = await eslint.lintText(tsExpression.tsExpression());
```
ueokande, line 33: To guard against unexpected errors, raise an error if the results contain non-fixable errors, such as syntax errors.

ueokande (changes requested):

packages/dts-gen/src/templates/template.ts
mshrtsr: fixed in 88447a72f89aadc64b3a2c6aa52ee5ee66d74bf2 .

packages/dts-gen/src/templates/template.ts
mshrtsr: fixed in https://github.com/kintone/js-sdk/pull/1266/commits/3cd1aa79ad96f980a7de385f11d2f37f6ccf93a7

packages/dts-gen/src/templates/template.ts
mshrtsr: fixed in https://github.com/kintone/js-sdk/pull/1266/commits/d6abfa9dc4a3557985c84711cafc344ed31a0d19

packages/dts-gen/src/templates/template.ts
mshrtsr: fixed in https://github.com/kintone/js-sdk/pull/1266/commits/3cd1aa79ad96f980a7de385f11d2f37f6ccf93a7
(If eslintResult does not have the `output` property, original code is used.)

packages/dts-gen/src/templates/template.ts
```diff
@@ -11,18 +12,47 @@ interface RenderInput {
...
   });
   const outputPath = path.join(process.cwd(), output);
 
-  if (!fs.existsSync(path.dirname(outputPath))) {
-    fs.mkdirSync(path.dirname(outputPath));
-  }
-
-  fs.writeFileSync(path.join(process.cwd(), output), prettySource);
+  await fs.promises.mkdir(path.dirname(outputPath), { recursive: true });
+  await fs.promises.writeFile(path.join(process.cwd(), output), prettySource);
```
ueokande, line 55:
```suggestion
  await fs.promises.writeFile(outputPath, prettySource);
```

PR #1349, opened 2022-02-22
fix(data-loader): export csv correctly when Table value is empty array

ksm-cy:
## Why

When a Table's value is an empty array, exporting in csv format results in an empty row, even though there're values in fields outside the Table that I want to export.
This happens as Table value was always supposed to be non-empty.
※Table with value of an empty array is rare but not impossible. One way of reproducing this would be adding a Table to a kintone APP with existing records - the Table value for those records would be `[]`.

## What

- [x] support CSV export when Table value is an empty array
- [x] fix tests

## How to test

```sh
$ yarn build
$ yarn lint && yarn test
```

```sh
$ yarn build
$ cd packages/data-loader
$ node lib/main.js export --format="csv" --app <appId>
```

## Checklist


mshrtsr (approved):
LGTM!! 💯

PR #1423, opened 2022-04-04
feat(data-loader): upsert-records-with-unique-field

kjmnmt:
## Why

Being able to upsert (update and add) records adds great value to data-loader as a data import tool, especially in cases where data comes from another system and is backed up to kintone.
This ensures that existing data is updated, while new data is imported correctly.

## What

We divide user data into two groups by comparing the value of "Key to Bulk Update" from the user data and the existing kintone records.
The group with matching key values becomes the "update group", meaning those matched existing records are updated accordingly. The other group containing keys that don't have a match - which means they are new keys - is the "add group", whose data is added to the app as new records.

## How to test

```
$ yarn build
$ yarn lint && yarn test
```

```
$ yarn build
$ cd packages/data-loader
$ node lib/main.js import --app <appId> --file-path <path-to-csv-file> --update-key <key-to-bulk-update>
```

## Checklist


shintaroNagata (approved):
👍

PR #1458, opened 2022-04-26
feat(data-loader): support specify character encoding when import

kjmnmt:
## Why

Because we want to import CSV files encoded with various encodings.
(Especially, for Shift JIS encoding in the Windows environment.)

## What

- [x] add `encoding` option to import sub-command

## How to test

```
$ yarn build
$ yarn lint && yarn test
```

```
$ yarn build
$ cd packages/data-loader
# Prepare 
$ node lib/main.js import --app <appId> --file-path <path-to-csv-file> --encoding sjis
```

## Checklist


packages/data-loader/src/controllers/import.ts
```diff
@@ -1,5 +1,5 @@
 import { buildRestAPIClient, RestAPIClientOptions } from "../api";
-import { readFile } from "../utils/file";
+import { ImportEncoding, readFile } from "../utils/file";
```
ksm-cy, line 2: `ImportEncoding`って名前が動詞に見えてしょうがなくてw（readFileと並んでいるからかな）
`SupportedEncoding`(`SupportedImportEncoding`)などはどうでしょうか？

packages/data-loader/src/commands/import.ts
```diff
@@ -96,6 +104,7 @@ export const handler = (args: Args) => {
     attachmentsDir: args["attachments-dir"],
     filePath: args["file-path"],
     updateKey: args["update-key"],
+    encoding: args.encoding,
```
ksm-cy, line 107: 細かいんですが、他と形式を揃えて`args["encoding"]`にしませんか

packages/data-loader/src/utils/file.ts
```diff
@@ -1,20 +1,27 @@
...
-  const stream = fs.createReadStream(filePath);
+  filePath: string,
+  encoding?: ImportEncoding
+) => Promise<{ content: string; format: string }> = async (
+  filePath,
+  encoding
+) => {
+  const stream = fs
+    .createReadStream(filePath)
+    .pipe(iconv.decodeStream(encoding || "utf8"));
```
ksm-cy, line 16: `encoding`のデフォルト値が設定されているなら、`|| "utf8"`部分がいらなくなります・・？

packages/data-loader/src/commands/import.ts
ksm-cy: Okay, it's ESLint who formats it like this 💯

packages/data-loader/src/utils/file.ts
kjmnmt: @ksm-cy 
Thanks for the comment!
The default with the `--encoding` option is "utf8", but without the option, it is undefined.

ueokande (approved):

PR #1495, opened 2022-05-12
feat(rest-api-client): add `httpsAgent` property to constructor options

mshrtsr:
## Why

fixes kintone/js-sdk-ja#24 .

Currently, rest-api-client generates HTTPS agent internally with user given options (`clientCertAuth`)
Sometimes users want to use a customized HTTPS agent; e.g. request over proxy server HTTP tunneling.

In this PR, we add `httpsAgent` property  to KintoneRestApiClient constructor options.

## What

- [x] add `httpsAgent` property
- [x] update README.md

## How to test

```
$ yarn build
$ yarn lint && yarn test
```

## Checklist


packages/rest-api-client/src/KintoneRequestConfigBuilder.ts
```diff
@@ -46,37 +48,44 @@ export class KintoneRequestConfigBuilder implements RequestConfigBuilder {
...
+    this.headers = this.buildHeaders({
+      basicAuth: options.basicAuth,
+      userAgent: options.userAgent,
+    });
+    if ("httpsAgent" in options) {
+      this.httpsAgent = options.httpsAgent;
+    }
+    if ("clientCertAuth" in options) {
+      this.clientCertAuth = options.clientCertAuth;
+    }
```
ueokande, line 87: I believe the method should throw an error when the user specifies both the parameter `clientCertAuth` and `httpAgent`.  The README says:

> When `httpsAgent` parameter is given, this parameter is *unavailable*.

But actually, the `httpsAgent` is *ignored* rather than *unavailable*.

packages/rest-api-client/src/KintoneRestAPIClient.ts
```diff
@@ -18,21 +19,25 @@ type Options = {
...
+        | {
+            pfx: Buffer;
+            password: string;
+          }
+        | {
+            pfxFilePath: string;
+            password: string;
+          };
+    }
+);
```
ueokande, line 40: This `Options` type can be simplified as the following:
```typescript
type Options = {
  baseUrl?: string;
  ....

  httpsAgent?: https.Agent,
  clientCertAuth?: 
          | {
            pfx: Buffer;
            password: string;
          }
        | {
            pfxFilePath: string;
            password: string;
          };
}
```

ueokande (changes requested):

ueokande (approved):

Issue #1299, opened 2022-01-17
plugin.zip are not generated when `webpack --watch`


shintaroNagata:
## Target Package

- @kintone/webpack-plugin-kintone-plugin v5.0.32

## Environment

- Operating System: macOS Big Sur
- Node.js version: v16.13.0

## Current Behavior

When I execute `webpack --watch`, my plugin codes are compiled successfully, but plugin.zip are not generated.
(When I give some changes to my codes, my codes are recompiled, and plugin.zip are generated correctly.)

## Expected Behavior

plugin.zip are always generated even I execute `webpack --watch`.

PR #1529, opened 2022-05-24
fix(webpack-plugin-kintone-plugin): generate plugin.zip when watch begins

mshrtsr:
## Why

fixes https://github.com/kintone/js-sdk/issues/1299.

When webpack launches with the `--watch` option, webpack-plugin-kintone-plugin does not create plugin.zip before file changes.

## What

- [x] fix plugin code
- [x] add test

## How to test

```shell
$ yarn install && yarn build
$ cd ./packages/webpack-plugin-kintone-plugin
$ yarn test
```

## Checklist


packages/webpack-plugin-kintone-plugin/src/index.ts
```diff
@@ -46,6 +47,12 @@ class KintonePlugin implements WebpackPluginInstance {
       }
       this.privateKey = fs.readFileSync(privateKeyPath, "utf-8");
       if (compiler.options.watch) {
+        compiler.hooks.afterEmit.tapPromise(this.name, async () => {
+          if (this.isFirstEmitting) {
```
shisama, line 51: Why is this `isFirstEmitting` flag necessary? Is the following code invalid?

```ts
compiler.hooks.afterEmit.tapPromise(this.name, () =>
  this.generatePlugin()
);
if (compiler.options.watch) {
  this.watchAssets();
}
```

mshrtsr: @shisama 
Thank you for the review!!

> Why is this isFirstEmitting flag necessary?

Because the function registered to `compiler.hooks.afterEmit.tapPromise` is triggered every time webpack emit assets.
In the case of your suggestion code, when src/customize.js is changed, webpack detects it and emits assets.
And then `generatePlugin()` is triggered and generates plugin.zip.
On the other hand, `watchAssets()` also detects the assets changes and generates plugin.zip.

I added a test case about it in https://github.com/kintone/js-sdk/pull/1529/commits/1ea0190641fa804e26ebfbf4b75f6b1997954667.

Is there any way to trigger only once on afterEmit?
I cloud not find it in the documentation.

mshrtsr:
TODO
- [ ] replace tests that call child process with calling webpack Node API
  - https://webpack.js.org/api/node/

packages/webpack-plugin-kintone-plugin/src/index.ts
```diff
@@ -46,6 +47,12 @@ class KintonePlugin implements WebpackPluginInstance {
       }
       this.privateKey = fs.readFileSync(privateKeyPath, "utf-8");
       if (compiler.options.watch) {
```
mshrtsr, line 49: use `compiler.watchMode`.

NOTE: Node API `webpack.Compiler.watch()` does not set `compiler.options.watch` to `true`.

sosukesuzuki (approved):
I think it's great we got rid of `watch.ts`! LGTM!

PR #1837, opened 2022-11-10
test(plugin-packer): update test case due to breaking changes of glob v8

hung-cybo:
## Why

There are breaking changes in glob v8, so we have to update the corresponding test case.

Ref: https://github.com/kintone/js-sdk/pull/1443#discussion_r1019055983

## What

- [x] test case which is using glob v8 is updated

## How to test

```
yarn build

yarn test
```

## Checklist


mshrtsr (approved):
LGTM!

PR #1943, opened 2023-03-02
fix: replace mime-types by mime package

hung-cybo:
## Why

Fix #1939 

## What

- Replace [mime-types](https://github.com/jshttp/mime-types) by [mime](https://github.com/topics/mine) package

## How to test

N/A

## Checklist

- [x] Read [CONTRIBUTING.md](https://github.com/kintone/js-sdk/blob/master/CONTRIBUTING.md)
- [ ] Updated documentation if it is required.
- [ ] Added tests if it is required.
- [x] Passed `yarn lint` and `yarn test` on the root directory.

PR #1945, opened 2023-03-03
test(rest-api-client): add tests for buildFormDataValue

mshrtsr:
## Why

In https://github.com/kintone/js-sdk/pull/1943, we fixed the bug of `buildFormDataValue()`.
It's better to have a small test to check auto-detection of the MIME type works correctly.
So that we can change our dependencies with the relief.

## What

- [x] add test for `buildFormDataValue()`
  - I changed the [test environment](https://jestjs.io/ja/docs/configuration#testenvironment-string) of `browser.test.ts`.

## How to test

```shell
yarn install
yarn build

yarn test
```

## Checklist


hung-cybo (approved):
LGTM

PR #2009, opened 2023-04-03
An error is caught, ends with exit status 1

the-red:
## Why

When an error is caught by the handler() function, the command exits with exit status 0, so I'm unable to do error handling from a parent process.

Specifically, I am calling kintone-dts-gen in a loop within my tool "Goqoo on kintone" as follows, and I want to judge if the dts file creation succeeded or failed.
https://github.com/goqoo-on-kintone/goqoo/blob/267c1fbbf87b80f52e4882a0955c8f10ee6e8a01/src/generator/dts/index.ts#L53-L84

## What

An error is caught, ends with exit status other than 0.
I returned 2 because there was already a process that returned 1, but it can be anything other than 2.

## How to test

```shell
$ kintone-dts-gen --base-url example.cybozu.com -u hoge -p fuga --app-id 0 # Failed
$ echo $? # Before:0 / After:2
```

## Checklist

### DIFF
--- a/packages/dts-gen/src/index.ts
+++ b/packages/dts-gen/src/index.ts
@@ -37,5 +37,5 @@ const handler = async () => {
 handler().catch((err) => {
   console.error(err);
   // eslint-disable-next-line no-process-exit
-  process.exit(2);
+  process.exit(1);
 });
--- a/packages/dts-gen/src/index.ts
+++ b/packages/dts-gen/src/index.ts
@@ -34,4 +34,8 @@ const handler = async () => {
   await TypeDefinitionTemplate.renderAsFile(args.output, input);
 };
 
-handler().catch((err) => console.error(err));
+handler().catch((err) => {
+  console.error(err);
+  // eslint-disable-next-line no-process-exit
+  process.exit(2);
+});


packages/dts-gen/src/index.ts
```diff
@@ -34,4 +34,8 @@ const handler = async () => {
   await TypeDefinitionTemplate.renderAsFile(args.output, input);
 };
 
-handler().catch((err) => console.error(err));
+handler().catch((err) => {
+  console.error(err);
+  // eslint-disable-next-line no-process-exit
+  process.exit(2);
```
mshrtsr, line 40:
```suggestion
  process.exit(1);
```

We need more time to consider the return values for each error.
Please keep using `1` as the exit status for all failures at this time.

mshrtsr (changes requested):
@the-red 
Thank you for your contribution!
There is a code that needs to be modified before merging.

packages/dts-gen/src/index.ts
the-red: @mshrtsr 
Thanks for the review.
I comitted your suggestion!

mshrtsr (approved):
LGTM!

the-red:
@mshrtsr 
Thank you!!!
Error handling in "Goqoo on kintone" now works correctly 😄 
https://github.com/goqoo-on-kintone/goqoo/commit/c5d19dd011935cd354538e53777e341bb2983127

PR #2017, opened 2023-04-05
ci: add `permissions` to required workflows

mshrtsr:
## Why

For security improvement, we will make the default workflow permission `read-only`.
In this PR, we update the `release` and `label` workflow because it requires additional permission.

## Checklist


hung-cybo (approved):
LGTM

PR #2018, opened 2023-04-05
ci: stop using local checksum of yamory script

mshrtsr:
## Why

Currently, we validate the yamory scan script with SHA512 checksum to detect falsification.
We store the checksum on the workflow’s environment variable because it's a possible situation where both script and checksum are falsified.
However, it makes CI fail when the yamory script is updated.
We cannot predict when the yamory script will be updated, So CI fails randomly in effect.

In https://github.com/kintone/cli-kintone/pull/295, we made the default workflow permission `read-only`.
So that yamory workflow can only read repository contents now.
Our repositories are public, so there is no risk in reading the repo.

Therefore, we can remove stored checksum and download checksum file on each CI run. 

## What

- remove `YAMORY_YARN_SCRIPT_SHA512` env variable

## How to test

See the result of the latest CI run.
https://github.com/kintone/js-sdk/actions/runs/4618278806/jobs/8165529845?pr=2018

## Checklist

### DIFF
--- a/.github/workflows/yamory.yml
+++ b/.github/workflows/yamory.yml
@@ -18,14 +18,12 @@ jobs:
         node-version: 16.x
         cache: yarn
     - name: download and validate Yamory script
-      env:
-        # https://cli.yamory.io/yamory-app-library-scan-scripts/yamory-yarn-sha512sum.txt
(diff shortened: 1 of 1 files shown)


hung-cybo (approved):
LGTM

PR #2102, opened 2023-05-23
fix(rest-api-client): make the `proxy` option of `KintoneRestAPIClient` accept false

ryounasso:
## Why

Currently, the `proxy` option of KintoneRestAPIClient accepts only an object to configure proxy.
https://github.com/kintone/js-sdk/blob/0cf33a539104decf5f4ecebe0ba2e8ca9d846c67/packages/rest-api-client/src/KintoneRestAPIClient.ts#L23

Sometimes, we want to disable the proxy, ignoring environment variables like `https_proxy`.

Axios accepts `false` to disable proxy.
> Use `false` to disable proxies, ignoring environment variables.
https://github.com/axios/axios#request-config

However, specifying `false` to KintoneRestAPIClient causes a type error currently.
https://github.com/kintone/cli-kintone/blob/d01a232d9297fc41cdd69a88a4370af75db40267/src/kintone/client.ts#L100

## What

- make the `proxy` option of `KintoneRestAPIClient` accept false
- update the description of proxy option

## How to test

```typescript
const client = new KintoneRestAPIClient({proxy:false})
```

## Checklist


packages/rest-api-client/README.md
```diff
@@ -105,7 +105,7 @@ client.record
...
-| proxy                      |                              Object                              |                             | **This parameter is available only in Node.js environment.**<br />If you use a proxy, please specify its configuration.                                                                                                                                                                                                                                      |
+| proxy                      |                        Object or `false`                         |                             | **This parameter is available only in Node.js environment.**<br />If you use a proxy, please specify its configuration.<br>To disable proxy ignoring environment variables, please specify `false`.                                                                                                                                                          |
```
hung-cybo, line 108: @ryounasso 

```
To disable proxy ignoring environment variables, please specify `false`
```

I see this sentence is unclear. So, my suggestion is:

```
To disable proxy and ignore proxy environment variables, please specify `false`.
```

Please consider it. Thank you.

packages/rest-api-client/src/http/HttpClientInterface.ts
```diff
@@ -23,14 +23,16 @@ export type Response<T = any> = {
...
+export type ProxyConfig =
+  | {
+      host: string;
+      port: number;
+      auth?: {
+        username: string;
+        password: string;
+      };
+    }
+  | false;
```
hung-cybo, line 35: @ryounasso 

I think we should add a unit test for this change to make sure the `KintoneRequestConfigBuilder` can receive the `proxy` property as `false`.

hung-cybo (commented):
@ryounasso 

I have two comments.
Please help to check it.
Thank you.

packages/rest-api-client/src/http/HttpClientInterface.ts
ryounasso: Thank you for your review comments.

Is the test to be written as follows correct?
```typescript
describe("type test", () => {
    it("should accept false value", () => {
      const proxyConfig:ProxyConfig = false;
      expect(proxyConfig).toBe(false);
    })
  })
```

hung-cybo: @ryounasso 

You can refer to this test case:

https://github.com/kintone/js-sdk/blame/6e6fe3728218e72aaa6a4a9ffbe0926468ffee29/packages/rest-api-client/src/__tests__/KintoneRequestConfigBuilder.test.ts#L369-L387

For the new test case, we need to make sure that when passing `proxy: false` to create a new instance of `KintoneRequestConfigBuilder`, we can receive the `proxy` property after calling `kintoneRequestConfigBuilder.build`, and its value must be `false`.

Sorry, feel free to let me know if It still makes you confused.
Thank you.

packages/rest-api-client/README.md
ryounasso: Fixed with 5666c6ce3ed5bccbc3a952d432745297043fd6f8

packages/rest-api-client/src/http/HttpClientInterface.ts
ryounasso: @hung-cybo 
Fixed with 825e52fc5ed637cfcb41c0d5bf95f41ccd517575

hung-cybo:
LGTM

PR #2137, opened 2023-06-08
fix(dts-gen): validate namespace and typename

hung-cybo:
## Why

An unclear message is shown when specifying an invalid `namespace` or `type-name` option.
For example:

```
$ kintone-dts-gen --base-url https://*****.cybozu-dev.com -u ****** -p ****** --app-id 1 --type-name SampleFields --namespace sample-fields
```

** Invalid `namespace`: `sample-fields`
** Valid `namespace`: `sampleFields`

Current error message:
```
Error: failed to fix lint errors on generated type definition file.
    at /Users/h000953/Development/js-sdk/packages/dts-gen/dist/templates/template.js:61:15
    at Generator.next (<anonymous>)
    at fulfilled (/Users/h000953/Development/js-sdk/packages/dts-gen/dist/templates/template.js:28:58)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

## What

- [x] Showing a detailed error message
- [x] Validate `namespace` and `type-name` option

## How to test

Run command

```
$ kintone-dts-gen --base-url https://*****.cybozu-dev.com -u ****** -p ****** --app-id 1 --type-name SampleFields --namespace sample-fields
```

## Checklist

### DIFF
--- a/packages/dts-gen/src/cli-parser.ts
+++ b/packages/dts-gen/src/cli-parser.ts
@@ -1,6 +1,7 @@
 import { Command } from "commander";
+import { validateArgs } from "./validators/args";
 
-interface ParsedArgs {
+export interface ParsedArgs {
   baseUrl: string;
   username: string | null;
   password: string | null;
@@ -82,6 +83,7 @@ export const parse = (argv: string[]): ParsedArgs => {
 
   const options = program.opts();
   const {
+    baseUrl,
     username,
     password,
     apiToken,
@@ -98,12 +100,7 @@ export const parse = (argv: string[]): ParsedArgs => {
     output,
   } = options;
 
-  const baseUrl = options.baseUrl;
-  if (baseUrl === null) {
-    throw new Error("--base-url (KINTONE_BASE_URL) must be specified");
-  }
-
-  return {
+  const parsedArgs: ParsedArgs = {
     baseUrl,
     username,
     password,
@@ -120,4 +117,8 @@ export const parse = (argv: string[]): ParsedArgs => {
     namespace,
     output,
   };
+
+  validateArgs(parsedArgs);
+
+  return parsedArgs;
 };
--- a/packages/dts-gen/src/templates/template.ts
+++ b/packages/dts-gen/src/templates/template.ts
@@ -3,14 +3,8 @@ import * as path from "path";
 import * as prettier from "prettier";
 import { ESLint } from "eslint";
 
-import type { FieldTypeGroups } from "../converters/fileldtype-converter";
+import type { RenderInput } from "../types/template";
 import { convertToTsExpression } from "./converter";
-
-interface RenderInput {
-  typeName: string;
-  namespace: string;
-  fieldTypeGroups: FieldTypeGroups;
-}
 
 const renderAsFile = async (output: string, renderInput: RenderInput) => {
   const tsExpression = convertToTsExpression(renderInput);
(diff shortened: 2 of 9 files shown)


packages/dts-gen/src/templates/template.ts
```diff
@@ -32,8 +26,15 @@ const renderAsFile = async (output: string, renderInput: RenderInput) => {
...
+    let errorMessage = "";
+    if (eslintResult.messages.length > 0) {
+      errorMessage = `Causes:`;
+      eslintResult.messages.forEach((error) => {
+        errorMessage += `\n- ${error.message}`;
+      });
+    }
     throw new Error(
-      "failed to fix lint errors on generated type definition file."
+      `Failed to fix lint errors on the generated type definition file.\n${errorMessage}`
```
mshrtsr, line 37: 👍

packages/dts-gen/src/validators/args.ts
```diff
@@ -0,0 +1,20 @@
+type Args = {
+  namespace?: string;
+  typeName?: string;
+};
+
+export const validateArgs = (args: Args) => {
+  if (args.namespace && !isValidIdentifier(args.namespace)) {
+    throw new Error("Invalid namespace option!");
```
mshrtsr, line 8: MUST: Please include a helpful message to help users to fix the namespace.
e.g., includes name convention in an error message.
 
```
Invalid namespace option.
The namespace and type-name convention:
- Starts with a letter (`a-z` or `A-Z`), underscore (`_`), or dollar sign (`$`).
- Can be followed by any alphanumeric, underscores, or dollar signs.
```

mshrtsr (changes requested):
Please update the error message. The rest of the code seems good!
I also commented on your refactoring.

packages/dts-gen/src/validators/__tests__/args.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    },
+  },
+];
+
+describe("validateInput", () => {
+  test.each(patterns)("$description", (pattern) => {
+    if (pattern.expected.failure !== undefined) {
+      return expect(() => {
+        validateArgs(pattern.input);
+      }).toThrow(pattern.expected.failure.errorMessage);
```
mshrtsr, line 90: MUST:
Because we added additional information to the error message, the `invalidNamespaceMessage` and `invalidTypeNameMessage` is incorrect now.
They must includes `identifierConventionMsg`.

However, CI passes in false negative.
According to the Jest document, `toThrow()` accepts not only string but also error objects.
> string: error message includes the substring
> error object: error message is equal to the message property of the object
https://jestjs.io/docs/expect#tothrowerror

It's safe to use an error object than a string.
```suggestion
      }).toThrow(new Error(pattern.expected.failure.errorMessage));
```

packages/dts-gen/src/cli-parser.ts
```diff
@@ -120,4 +117,8 @@ export const parse = (argv: string[]): ParsedArgs => {
     namespace,
     output,
   };
+
+  validateArgs(parsedArgs);
```
mshrtsr, line 121: It's an excellent decision to validate the argument inside parser 👍 

IMO: We can place the validator inside cli-parser. (Just IMO)

mshrtsr (changes requested):
The test seems to pass in false-negative.
Please check my comment:)

mshrtsr (approved):
LGTM!!

PR #2158, opened 2023-06-29
fix(dts-gen): remove unnecessary `exclude` from tsconfig

mshrtsr:
## Why

Currently, the document of the dts-gen guide to add `files`, `include`, and `exclude` to tsconfig.json.
However, `exclude` is to filter out some files from the result of `include`, so that is unnecessary in this case.

> exclude only changes which files are included as a result of the include setting.
https://www.typescriptlang.org/tsconfig#include
https://www.typescriptlang.org/tsconfig#exclude

## What

- remove `exclude` from tsconfig of the document

## How to test

N/A

## Checklist

### DIFF
--- a/packages/dts-gen/docs/how-to-use.md
+++ b/packages/dts-gen/docs/how-to-use.md
@@ -123,9 +123,6 @@ npx tsc --init
   ],
   "include": [
     "src/**/*"
-  ],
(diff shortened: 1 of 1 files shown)


hung-cybo (approved):
LGTM

PR #2159, opened 2023-06-29
fix(create-plugin): fix `include` of tsconfig

mshrtsr:
## Why

The `include` of the tsconfig of the create-plugin has typo to `indlude`.

## What

- fix type `indlude` to `include`
- remove `exclude` that is unnecessary

## How to test

N/A

## Checklist


hung-cybo (approved):
LGTM

PR #2181, opened 2023-07-14
feat(rest-api-client): adding socket timeout option

hung-cybo:
## Why

When there is a network issue, the user should get a notice about the error earlier.

## What

- [x] In the Node.js environment, allow the user to specify the socket timeout when initializing KintoneRestAPIClient

## How to test

N/A

## Checklist


packages/rest-api-client/src/platform/index.ts
```diff
@@ -6,6 +6,7 @@ type PlatformDeps = {
   getRequestToken: () => Promise<string>;
   getDefaultAuth: () => DiscriminatedAuth;
   buildPlatformDependentConfig: (params: object) => object;
+  buildPlatformTimeout: (params: { socketTimeout?: number }) => object;
```
mshrtsr, line 9: IMO: We can merge `buildPlatformTimeout()` to `buildPlatformDependentConfig()`.

packages/rest-api-client/src/__tests__/KintoneRequestConfigBuilder.test.ts
```diff
@@ -366,7 +366,7 @@ describe("options", () => {
     });
   });
 
-  it("should accept false when specify false to proxy opption", async () => {
+  it("should accept false when specify false to proxy option", async () => {
```
mshrtsr, line 369: 😄 👍

packages/rest-api-client/src/KintoneRestAPIClient.ts
```diff
@@ -142,3 +144,14 @@ const validateGuestSpaceId = (guestSpaceId: Options["guestSpaceId"]) => {
...
+const validateSocketTimeout = (socketTimeout: Options["socketTimeout"]) => {
+  if (socketTimeout === undefined) {
+    return;
+  }
+
+  const number = parseFloat(socketTimeout.toString());
+  if (isNaN(number) || socketTimeout < 0) {
+    throw new Error(`Invalid socketTimeout. Must be a positive number.`);
+  }
+};
```
mshrtsr, line 157: Can I ask you what you validate with this block?
I thought we only needed to check `socketTimeout < 0` to check if positive number.

Is it for non-TypeScript users who can specify non-number values?
If so, I think it's great! But there are some points that can be improved.
- MUST: Please use check `number < 0` rather than the original `socketTimeout` (because it can be non-number)
- IMO: add `string` to the type of input to make it clear what we want to guard with this validation (or use `unknown` to make it more strict)

```suggestion
const validateSocketTimeout = (socketTimeout: Options["socketTimeout"] | string) => {
  if (socketTimeout === undefined) {
    return;
  }

  const number = parseFloat(socketTimeout.toString());
  if (isNaN(number) || number < 0) {
    throw new Error(`Invalid socketTimeout. Must be a positive number.`);
  }
};
```

mshrtsr (commented):
There are no critical problems!
I left some comments requesting small changes.
Please check it out.

packages/rest-api-client/src/KintoneRestAPIClient.ts
hung-cybo: @mshrtsr 

> Is it for non-TypeScript users who can specify non-number values?

Yes, correct. We should validate the number format too.
When I check with non-Typescript example and specify `socketTimeout` as a string, the process will hang and there is no notice.

For the negative number, there also has an error message about this. But, It's not clear, so It's hard for the user to know what going on, so I added the validation for the positive number as well.

```
node:internal/timers:396
    throw new ERR_OUT_OF_RANGE(name, 'a non-negative finite number', msecs);
    ^

RangeError [ERR_OUT_OF_RANGE]: The value of "msecs" is out of range. It must be a non-negative finite number. Received -100
```

> Please use check number < 0 rather than the original socketTimeout

Oh, this is my mistake. Thank you.
It should be `number < 0`, not `socketTimeout`

> add string to the type of input to make it clear what we want to guard with this validation

Great. I will update it.

packages/rest-api-client/src/platform/index.ts
hung-cybo: @mshrtsr 

Yes, you're right. we can merge them in `buildPlatformDependentConfig`.
I refactored it.

packages/rest-api-client/src/platform/node.ts
```diff
@@ -56,6 +73,14 @@ export const buildPlatformDependentConfig = ({
   return {};
 };
 
+export const buildTimeoutConfig = (params: { socketTimeout?: number }) => {
```
mshrtsr, line 76: IMO: Now, this function doesn't need to be exported.

```suggestion
const buildTimeoutConfig = (params: { socketTimeout?: number }) => {
```

mshrtsr (approved):
@hung-cybo 
Approved!!
I left an additional comment. I leave it to you to fix or not.

PR #2191, opened 2023-07-23
chore: apply prettier for yml, json, and md files

hung-cybo:
## Why

YAML, Json and MD files should be formatted automatically so that we don't need to care about formatting

## What

- [x] yml, json and md files are formatted
- [x] fix and lint scripts also apply for yml, json and md files

## How to test

`pnpm lint`

## Checklist


mshrtsr (changes requested):
I left some comments to request changes.

---

In addition, please specify `root: true` to .eslintconfig in root directory.
https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy

mshrtsr (approved):
LGTM!

Issue #1939, opened 2023-02-27
rest-api-client with webpack outputs `Module not found: Error: Can't resolve 'path'`


ueokande:
## Target Package

- @kintone/rest-api-client 3.3.0

## Environment

- Operating System: macOS 13.2
- Browser:
- Node.js version: 16.16.0

## Current Behavior

https://github.com/kintone/js-sdk-ja/issues/36

Webpack outputs the following error on @kintone/rest-api-client 3.2.0
```
ERROR in ./node_modules/mime-types/index.js 16:14-37
Module not found: Error: Can't resolve 'path' in '/Users/xxx/yyy/node_modules/mime-types'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }
```

This occur occurs by https://github.com/kintone/js-sdk/pull/1894.  This change have added a new dependency [mime-types](https://www.npmjs.com/package/mime-types), which depends on Node.js path package ([see](https://github.com/jshttp/mime-types/blob/master/index.js#L16)).  This problem affects webpack users.  For [UMD](https://unpkg.com/browse/@kintone/rest-api-client@3.2.0/), it works well on the browser because it contains path in the content.

We will fix the problem in the future version.

## Expected Behavior

Webpack build successfully on latest rest-api-client without build configuration changes.

PR #2381, opened 2023-11-06
build: detect unavailable deps for NodeJS

tuanphamcybozu:
## Why

We want to prevent using environment-specific APIs because rest-api-client is an isomorphic library.
We caused an error, including Node.js API in the browser build.
https://github.com/kintone/js-sdk/issues/1939

## What

In this PR, we support detecting if our code uses the browser-only function/package in the build for Node.js.

- Created multiple tsconfig files for building and type-checking
  - `tsconfig.json`: A root config (We use [Solution Style](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#support-for-solution-style-tsconfigjson-files))
    - `tsconfig.build.json`: A config to build CJS output
    - `tsconfig.build.esm.json`: A config to build ESM output
    - `tsconfig.typecheck.json`: A config to run type-check for all the source files
    - `tsconfig.typecheck.node.json`: A config to run type-check for the source files for the Node.js environment
- Excluding test files from the build
- Update `@types/node` to v18, the oldest Node version we support
- Refactor global object declarations

## How to test
```
$ cd packages/rest-api-client
$ pnpm build
```

We checked that CI failed by accessing DOM API on files for Node.js.
https://github.com/kintone/js-sdk/actions/runs/6781745569

## Checklist

### DIFF
--- a/packages/rest-api-client/package.json
+++ b/packages/rest-api-client/package.json
@@ -65,7 +65,12 @@
     "rollup-plugin-ecma-version-validator": "^0.2.13",
     "rollup-plugin-license": "^3.2.0",
     "rollup-plugin-node-globals": "^1.4.0",
-    "rollup-plugin-polyfill-node": "^0.12.0"
+    "rollup-plugin-polyfill-node": "^0.12.0",
+    "webpack": "^5.89.0",
+    "webpack-cli": "^5.1.4",
+    "babel-loader": "^9.1.3",
+    "vite": "^4.4.5",
+    "rimraf": "^5.0.5"
   },
   "dependencies": {
     "axios": "^1.6.0",
--- a/packages/rest-api-client/package.json
+++ b/packages/rest-api-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kintone/rest-api-client",
-  "version": "5.0.2",
+  "version": "5.0.3",
   "publishConfig": {
     "access": "public"
   },
@@ -64,15 +64,15 @@
     "@types/js-base64": "^3.0.0",
     "@types/mime": "^3.0.3",
     "@types/qs": "^6.9.9",
-    "rollup": "^4.1.4",
+    "rollup": "^4.2.0",
     "rollup-plugin-ecma-version-validator": "^0.2.13",
     "rollup-plugin-license": "^3.2.0",
     "rollup-plugin-node-globals": "^1.4.0",
     "rollup-plugin-polyfill-node": "^0.12.0"
   },
   "dependencies": {
-    "axios": "^1.5.1",
-    "core-js": "^3.33.1",
+    "axios": "^1.6.0",
+    "core-js": "^3.33.2",
     "form-data": "^4.0.0",
     "js-base64": "^3.7.5",
     "mime": "^3.0.0",
--- a/packages/rest-api-client/package.json
+++ b/packages/rest-api-client/package.json
@@ -15,9 +15,7 @@
   "types": "lib/src/index.d.ts",
   "scripts": {
     "prebuild": "pnpm clean",
-    "build": "run-p compile:*",
-    "compile:browser": "tsc --build --force tsconfig.json",
-    "compile:node": "tsc --build --force tsconfig.node.json",
+    "build": "tsc --build --force tsconfig.build.json",
     "postbuild": "run-p build:*",
     "lint": "eslint 'src/**/*.ts' --max-warnings 0",
     "prepublishOnly": "run-p build:umd_*",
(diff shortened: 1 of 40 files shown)


tuanphamcybozu:
This code for testing:
https://github.com/kintone/js-sdk/pull/2381/files#diff-9324125a349b923ca3bf2ff7a4597d3195c66af433a0debd4bbe193200ca0a0aR9-R11

I will remove it when the review is completed.

packages/rest-api-client/package.json
```diff
@@ -15,7 +15,9 @@
   "types": "lib/src/index.d.ts",
   "scripts": {
     "prebuild": "pnpm clean",
-    "build": "tsc --build --force",
+    "build": "run-p compile:*",
+    "compile:browser": "tsc --build --force tsconfig.json",
+    "compile:node": "tsc --build --force tsconfig.node.json",
```
mshrtsr, line 20: Because all the source files are transpiled with `tsconfig.json`, we don't need to generate output with `tsconfig.node.json`.
(The `src/**/node.ts` and `UnsupportedPlatformError.ts` are generated twice)

The `tsconfig.node.json` will be only for type checking.

```suggestion
    "build": "run-p compile typecheck",
    "compile": "tsc --build --force tsconfig.json",
    "typecheck": "tsc --build --force tsconfig.node.json",
```

In addition, we want to find browser-only dependencies not only on `node.ts` and `UnsupportedPlatformError.ts` but also all the files excluding `browser.ts` and its related files.

mshrtsr (changes requested):

tuanphamcybozu:
> `tsconfig.typecheck.json`: A config to run type-check for the source files for the Node.js environment
Excluding test files from the build
`tsconfig.typecheck.json` should be `tsconfig.typecheck.node.json`, right?

mshrtsr:
@tuanphamcybozu 

> > `tsconfig.typecheck.json`: A config to run type-check for the source files for the Node.js environment
> > Excluding test files from the build
> 
> `tsconfig.typecheck.json` should be `tsconfig.typecheck.node.json`, right?

Exactly! I fixed the PR body!

packages/rest-api-client/src/platform/browser.ts
```diff
@@ -42,7 +42,7 @@ export const buildHeaders = () => {
 };
 
 export const buildFormDataValue = (data: unknown, fileName?: string) => {
-  const options: BlobOptions = {};
+  const options: BlobPropertyBag = {};
```
tuanphamcybozu, line 45: @mshrtsr 
What happens if we still use BlobOptions?

packages/rest-api-client/tsconfig.typecheck.node.json
```diff
@@ -0,0 +1,24 @@
+{
+  "extends": "./tsconfig.typecheck.json",
+  "compilerOptions": {
```
tuanphamcybozu, line 3: When extends `./tsconfig.typecheck.json` and override `compilerOptions`,
`noEmit: true` is still applied for `tsconfig.typecheck.node.json`, is it right?

packages/rest-api-client/tsconfig.json
```diff
@@ -2,15 +2,36 @@
...
+    },
+    {
+      "path": "./tsconfig.build.esm.json"
+    },
+    {
+      "path": "./tsconfig.typecheck.json"
+    },
+    {
+      "path": "./tsconfig.typecheck.node.json"
+    }
```
tuanphamcybozu, line 35: How do you think in case of type-checking error,
if we change the order of `typechecks` before `build`, can the process be faster?

Like this:
```
"references": [
   {
      "path": "./tsconfig.typecheck.json"
    },
    {
      "path": "./tsconfig.typecheck.node.json"
    }
    {
      "path": "./tsconfig.build.json"
    },
    {
      "path": "./tsconfig.build.esm.json"
    },
]
```

packages/rest-api-client/src/platform/browser.ts
mshrtsr: @tuanphamcybozu 
The `BlobOptions` is the custom type defined in `src/types/global/index.d.ts`.
The `Blob` and `BlobPart` types are also defined there.
https://github.com/kintone/js-sdk/blob/b6622d63083d0f23287ee97a75e052234d776b85/packages/rest-api-client/types/global/index.d.ts

Actually, this Blob is the Blob of DOM API, and we can use the types in `lib.dom.d.ts`.
https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob
Using our custom type might cause a mismatch between our code and the actual DOM API.
So I removed these custom types and started using types from `lib.dom.d.ts`.

packages/rest-api-client/tsconfig.typecheck.node.json
mshrtsr: @tuanphamcybozu 
I couldn't find this behavior in the documentation.
https://www.typescriptlang.org/tsconfig#extends

But now, yes.
As you can see for the `compilerOptions.lib`, each property of the `compilerOptions` object is inherited.

packages/rest-api-client/tsconfig.json
mshrtsr: @tuanphamcybozu 
I checked swapping the order, but there was no significant difference in elapsed time.
Even if typecheck is failed with `./tsconfig.typecheck.node.json`, the build for `./tsconfig.build.json` will run and generate output.

tuanphamcybozu (commented):
LGTM

mshrtsr (approved):

PR #2497, opened 2024-01-12
chore(rest-api-client): Add warning message in Readme of getAllRecordsWithOffset

minh-nguyen1985:
## Why

The error occurs when using getAllRecordsWithOffset with an application containing more than 10,000 records

## What

Add warning message in Readme of getAllRecordsWithOffset

## How to test

N/A

## Checklist

### DIFF
--- a/packages/rest-api-client/docs/record.md
+++ b/packages/rest-api-client/docs/record.md
@@ -205,7 +205,7 @@ The returned array is sorted by `id` in ascending order.
 Retrieves details of all records from an app by specifying the app ID, fields, condition, and sort order.
 This method can retrieve the records exceeding the [REST API limitation](https://kintone.dev/en/docs/kintone/rest-api/overview/kintone-rest-api-overview/#limitations).
 
-:warning: **If you'd like to get over 10000 records, please consider using [getAllRecordsWithCursor](#getAllRecordsWithCursor) instead.**
+:warning: **An error will occur if the app has over 10.000 records. If you'd like to get over 10.000 records, please consider using [getAllRecordsWithCursor](#getAllRecordsWithCursor) instead.**
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


packages/rest-api-client/docs/record.md
```diff
@@ -205,6 +205,8 @@ The returned array is sorted by `id` in ascending order.
 Retrieves details of all records from an app by specifying the app ID, fields, condition, and sort order.
 This method can retrieve the records exceeding the [REST API limitation](https://kintone.dev/en/docs/kintone/rest-api/overview/kintone-rest-api-overview/#limitations).
 
+:warning: **If you'd like to get over 10000 records, please consider using [getAllRecordsWithCursor](#getAllRecordsWithCursor) instead.**
```
hung-cybo, line 208: @minh-nguyen1985 

As the description of this API `Retrieves details of all records from an app`, It means that when using this API, It will retrieve all records of the specific app. The user can not decide the number of records that will be returned.

And, the fact is the error will occur if the app has over 10.000 records. So, I think we can write it to the warning message. It's easier to understand for the user.

Please consider it.
Thank you.

hung-cybo (approved):
LGTM

packages/rest-api-client/docs/record.md
```diff
@@ -205,6 +205,8 @@ The returned array is sorted by `id` in ascending order.
 Retrieves details of all records from an app by specifying the app ID, fields, condition, and sort order.
 This method can retrieve the records exceeding the [REST API limitation](https://kintone.dev/en/docs/kintone/rest-api/overview/kintone-rest-api-overview/#limitations).
 
+:warning: **If the app has over 10.000 records, please consider using [getAllRecordsWithCursor](#getAllRecordsWithCursor) instead.**
```
mshrtsr, line 208: NITS: We use American English in our repository, so please use commas for the three digit separator.

```suggestion
:warning: **If the app has over 10,000 records, please consider using [getAllRecordsWithCursor](#getAllRecordsWithCursor) instead.**
```

mshrtsr (changes requested):

mshrtsr (approved):
LGTM!!

PR #2512, opened 2024-01-18
fix(dts-gen): adding type NULL to the error property

hung-cybo:
## Why

Missing the type NULL on the error property

## What

- Adding the type NULL on the error property

## How to test

N/A

## Checklist


mshrtsr (approved):
LGTM!

PR #2513, opened 2024-01-18
chore!(create-plugin): remove jQuery from the minimum template

tuanphamcybozu:
## Why
- jQuery is the old technology, so we don't recommend using jQuery in new projects anymore.
- We will stop using jQuery in our minimum template.

## What
- Remove jQuery from the minimum template

## How to test
- The create-plugin can generate a plugin project with an updated minimum template (without jQuery code inside)

## Checklist

### DIFF
--- a/packages/create-plugin/src/__tests__/fixtures/manifest.json
+++ b/packages/create-plugin/src/__tests__/fixtures/manifest.json
@@ -3,7 +3,7 @@
   "version": 1,
   "type": "APP",
   "desktop": {
-    "js": ["https://js.cybozu.com/jquery/3.3.1/jquery.min.js", "js/desktop.js"],
+    "js": ["js/desktop.js"],
     "css": ["css/51-modern-default.css", "css/desktop.css"]
   },
   "icon": "image/icon.png",
@@ -14,11 +14,11 @@
     "en": "kintone-plugin-demo"
   },
   "mobile": {
-    "js": ["https://js.cybozu.com/jquery/3.3.1/jquery.min.js", "js/mobile.js"]
+    "js": ["js/mobile.js"]
   },
   "config": {
     "html": "html/config.html",
-    "js": ["https://js.cybozu.com/jquery/3.3.1/jquery.min.js", "js/config.js"],
+    "js": ["js/config.js"],
     "css": ["css/51-modern-default.css", "css/config.css"],
     "required_params": []
   }
--- a/packages/create-plugin/src/manifest.ts
+++ b/packages/create-plugin/src/manifest.ts
@@ -3,20 +3,18 @@
 import type { Answers } from "inquirer";
 import type { TemplateType } from "./template";
 
-const jQueryURL = "https://js.cybozu.com/jquery/3.3.1/jquery.min.js";
-
 const minimumManifest = {
   manifest_version: 1,
   version: 1,
   type: "APP",
   desktop: {
-    js: [jQueryURL, "js/desktop.js"],
+    js: ["js/desktop.js"],
     css: ["css/51-modern-default.css", "css/desktop.css"],
   },
   icon: "image/icon.png",
   config: {
     html: "html/config.html",
-    js: [jQueryURL, "js/config.js"],
+    js: ["js/config.js"],
     css: ["css/51-modern-default.css", "css/config.css"],
     required_params: ["message"],
   },
@@ -121,10 +119,7 @@ export const buildManifest = (
       ...manifest,
       ...{
         mobile: {
-          js:
-            templateType === "minimum"
-              ? [jQueryURL, "js/mobile.js"]
-              : ["js/mobile.js"],
+          js: templateType === "minimum" ? ["js/mobile.js"] : ["js/mobile.js"],
           css: ["css/mobile.css"],
         },
       },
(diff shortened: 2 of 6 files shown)


packages/create-plugin/src/__tests__/fixtures/manifest.json
```diff
@@ -3,7 +3,7 @@
   "version": 1,
```
hung-cybo, line 3: @tuanphamcybozu 

I think this PR should be `feat`, not `chore`, because we drop the support of jQuery, It means that we have changed the behavior of this package.

Besides, when using `feat`, the user will know about this change via changelog.

packages/create-plugin/src/manifest.ts
```diff
@@ -121,10 +119,7 @@ export const buildManifest = (
       ...manifest,
       ...{
         mobile: {
-          js:
-            templateType === "minimum"
-              ? [jQueryURL, "js/mobile.js"]
-              : ["js/mobile.js"],
+          js: templateType === "minimum" ? ["js/mobile.js"] : ["js/mobile.js"],
```
hung-cybo, line 122: @tuanphamcybozu 

It's the same logic. Please remove the ternary operator.

packages/create-plugin/templates/minimum/src/js/config.js
```diff
@@ -1,27 +1,25 @@
-jQuery.noConflict();
```
hung-cybo, line 1: @tuanphamcybozu 

After creating a new plugin with the modern template, when accessing the plugin config page, an error is thrown.

`Uncaught Error: Required elements do not exist.`

Because `$form.length` or `$form.on` are syntaxes of jQuery.
Could you please check it?
Thank you.

hung-cybo (commented):
I left some comments.
Could you please check them?
Thank you.

packages/create-plugin/templates/minimum/src/js/config.js
```diff
@@ -1,27 +1,27 @@
...
+(function (PLUGIN_ID) {
+  "use strict";
 
-(function($, PLUGIN_ID) {
-  'use strict';
-
-  var $form = $('.js-submit-settings');
-  var $cancelButton = $('.js-cancel-button');
-  var $message = $('.js-text-message');
+  var $form = document.querySelectorAll(".js-submit-settings");
```
hung-cybo, line 4: @tuanphamcybozu 

Because we only get the first element, so I think we can use [querySelector()](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) instead.
Please consider it.
Thank you.

hung-cybo (approved):
LGTM

packages/create-plugin/templates/minimum/src/js/config.js
```diff
@@ -1,27 +1,25 @@
...
-  'use strict';
-
-  var $form = $('.js-submit-settings');
-  var $cancelButton = $('.js-cancel-button');
-  var $message = $('.js-text-message');
-  if (!($form.length > 0 && $cancelButton.length > 0 && $message.length > 0)) {
-    throw new Error('Required elements do not exist.');
+  var $form = document.querySelector(".js-submit-settings");
+  var $cancelButton = document.querySelector(".js-cancel-button");
+  var $message = document.querySelector(".js-text-message");
```
mshrtsr, line 6: Adding `$` as the prefix of variables is a custom of jQuery.
Please remove that prefix.

It's common to name Element as `el` now. (`el`, `formEl`, `cancelButtonEl`)
https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#examples

mshrtsr: Please use `const/let`, not `var`.
I expect you to replace these legacy/dangerous code in [the Boy Scout way](https://www.linkedin.com/pulse/leaving-code-cleaner-than-you-found-boy-scout-rule-software-chacko/).

mshrtsr (changes requested):
I added some change requests.

tuanphamcybozu:
- Run test job is failed after update var to const/let 
  Because '@cybozu/eslint-config/presets/kintone-customize-es5' is in eslint configuration
- WA: 
  Remove es5 configure and add remain configuration in extends property of eslint congifuration.

packages/create-plugin/templates/minimum/.eslintrc.js
```diff
@@ -1,5 +1,13 @@
...
+    '@cybozu/eslint-config/globals/kintone.js',
+    '@cybozu/eslint-config/lib/base.js',
+    '@cybozu/eslint-config/lib/kintone.js',
+  ],
+  rules: {
+    'prettier/prettier': ['error', { singleQuote: true }],
+    'space-before-function-paren': 0,
+    'object-curly-spacing': 0,
+  },
+  plugins: ['eslint-plugin-prettier'],
```
mshrtsr, line 12: We can use `prettier.js` to solve the CI error.
https://github.com/kintone/js-sdk/actions/runs/7635159605/job/20800124180#step:8:136

```suggestion
  extends: [
    '@cybozu/eslint-config/globals/kintone.js',
    '@cybozu/eslint-config/lib/base.js',
    '@cybozu/eslint-config/lib/kintone.js',
    '@cybozu/eslint-config/lib/prettier.js',
  ],
  rules: {
    'prettier/prettier': ['error', { singleQuote: true }],
    'space-before-function-paren': 0,
    'object-curly-spacing': 0,
  },
```

mshrtsr (approved):
LGTM!

PR #2519, opened 2024-01-22
fix(dts-gen): remove unavailable property

hung-cybo:
## Why

The `error` property is not available on the below fields ([ref](https://kintone.dev/en/docs/kintone/js-api/events/event-object-actions/)).
- Record number
- Created datetime
- Updated datetime

## What

- The `error` property is removed for the below fields
  - Record number
  - Created datetime
  - Updated datetime

## Checklist


mshrtsr (approved):
LGTM!

PR #2605, opened 2024-03-06
feat(rest-api-client): add `space.addThreadComment()` method

tuanphamcybozu:
## Why
- Support `addThreadComment` method to be able to add a comment to a thread of a space.

## What
- Add new method `addThreadComment`
- Add unit test
- Add demo script

## How to test
- Check test spec

## Checklist

- [X] Added tests if it is required.

### DIFF
--- a/packages/rest-api-client/src/client/types/space/index.ts
+++ b/packages/rest-api-client/src/client/types/space/index.ts
@@ -32,21 +32,20 @@ export type Space = {
   fixedMember: boolean;
 };
 
+type FileComment = {
+  fileKey: string;
+  width?: string | number;
+};
+
 export type CommentWithText = {
   text: string;
-  files?: Array<{
-    fileKey: string;
-    width?: string | number;
-  }>;
+  files?: FileComment[];
   mentions?: Entity[];
 };
 
 export type CommentWithFiles = {
   text?: string;
-  files: Array<{
-    fileKey: string;
-    width?: string | number;
-  }>;
+  files: FileComment[];
   mentions?: Entity[];
 };
 
--- a/packages/rest-api-client/src/client/types/space/index.ts
+++ b/packages/rest-api-client/src/client/types/space/index.ts
@@ -1,4 +1,5 @@
 import type { App } from "../app";
+import type { Entity } from "../entity";
 
 type AttachedApp = Pick<
   App,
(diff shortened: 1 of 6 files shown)


packages/rest-api-client/src/client/types/space/index.ts
```diff
@@ -30,3 +31,27 @@ export type Space = {
...
+  text?: string;
+  files: Array<{
+    fileKey: string;
+    width?: string | number;
+  }>;
+  mentions?: Entity[];
+};
+
+export type ThreadComment = {
+  space: string;
```
hung-cybo, line 54: Should type of `SpaceID`

packages/rest-api-client/src/client/types/space/index.ts
```diff
@@ -30,3 +31,27 @@ export type Space = {
...
+  thread: string;
```
hung-cybo, line 55: Should type of `ThreadID`
`string | number`

hung-cybo (commented):
I have some comments.
Please check it.
Thank you.

packages/rest-api-client/src/client/types/space/index.ts
```diff
@@ -30,3 +31,27 @@ export type Space = {
   attachedApps: AttachedApp[];
   fixedMember: boolean;
 };
+
+export type CommentWithText = {
+  text: string;
+  files?: Array<{
```
hung-cybo, line 37: The definition type of `fileds` is duplicated.

My suggestion: create a new type

```
type FileComment = {
    fileKey: string;
    width?: string | number;
}
export type CommentWithText = {
  text: string;
  files?: Array<FileComment>;
  mentions?: Entity[];
};
export type CommentWithFiles = {
  text?: string;
  files: Array<FileComment>;
  mentions?: Entity[];
};
```

packages/rest-api-client/src/client/types/space/index.ts
tuanphamcybozu: @hung-cybo 
I declared the name for corresponding to kintone REST API. 
https://kintone.dev/en/docs/kintone/rest-api/spaces/add-thread-comment/

So it's better to make parameter same between rest-api-client and kintone REST API.
How do you think?

About type of id, @tasshi-me said it should be string, instead of string or number.
Ref:
https://github.com/kintone/js-sdk/blob/0aaec0d726a044fdd3dff4853b432ebb6584f81a/packages/rest-api-client/src/client/types/space/index.ts#L18

tuanphamcybozu: after daily diccussion, I decided to update the type of ID to string | number
. Sorry for mismatch info.

hung-cybo (approved):
LGTM

PR #2718, opened 2024-04-22
test(webpack-plugin-kintone-plugin): fix failed test on windows

tuanphamcybozu:
## Why
- Fix fail test on Windows
- Following the 18.20.2/20.12.2 node.js releases, it is no longer valid to call child_process.spawn with a .cmd or .bat file without using the shell: true option.
Ref: https://nodejs.org/en/blog/release/v18.20.2

## What
- Add `shell: true` to `spawnSync` configuration if OS is Windows.

## How to test
`pnpm test`

## Checklist


hung-cybo (approved):
LGTM

PR #2720, opened 2024-04-23
test: add test case for JsSdkTest-1

tuanphamcybozu:
## Why
- Increase the reliability of create-plugin package.

## What
- Add test case [JsSdkTest-1](https://bozuman.cybozu.com/k/84147/show#record=1)

## How to test
`pnpm test:e2e`

## Checklist


packages/create-plugin/__e2e__/utils/helper.ts
```diff
@@ -0,0 +1,12 @@
+export const generateRandomString = (length: number) => {
```
tasshi-me, line 1: I couldn't find usages of this function.
Is it for future implementation?

packages/create-plugin/package.json
```diff
@@ -53,7 +53,8 @@
     "e2e": "cross-env NODE_ENV=e2e jest --testPathPattern=/__tests__/e2e\\.test\\.ts$",
     "test": "pnpm unittest",
     "test:all": "run-p test e2e",
```
tasshi-me, line 55: Don't we need to add `test:e2e` to `test:all`?

tasshi-me (approved):

packages/create-plugin/__e2e__/utils/executeCommand.ts
```diff
@@ -0,0 +1,145 @@
...
+  input
+    .replace(/\$\$[a-zA-Z0-9_]+/g, processEnvReplacer)
+    .replace(/\$[a-zA-Z0-9_]+/g, inputEnvReplacer(envVars));
+
+export const executeCommandWithInteractiveInput = async (
+  command: string,
+  workingDir: string,
+  outputDir: string,
+  questionsInput: QuestionInput[],
+  commandArguments?: string,
```
tasshi-me, line 69: Apply [RORO](https://www.freecodecamp.org/news/elegant-patterns-in-modern-javascript-roro-be01e7669cbd/) for functions with such a lot of inputs.

tasshi-me (changes requested):

packages/create-plugin/__e2e__/utils/helper.ts
hung-cybo: Yes, It will be used in the next test cases.

packages/create-plugin/package.json
hung-cybo: Yes. I think we can add `test:e2e` to `test:all`. But, let's do it in another PBI (Run e2e test on GitHub Actions).
Because if we add it now, the new test case of this PR also run on CI.
(It's out of scope of the PR)

tasshi-me: OK, I see!

PR #2810, opened 2024-06-10
test(create-plugin): optimizing e2e test

hung-cybo:
## Why

- The test cannot exit when exceeds the Jest timeout due to the spawn process.
- The execution time is too high.

## What

- Teardown spawn process when a test case is finished
- Reduce the execution time

## How to test

```
pnpm test:e2e
```

## Checklist

### DIFF
--- a/packages/create-plugin/__e2e__/utils/CreatePlugin.ts
+++ b/packages/create-plugin/__e2e__/utils/CreatePlugin.ts
@@ -25,6 +25,7 @@ export class CreatePlugin {
   private currentStep: number = 0;
   private stdout: string = "";
   private stderr: string = "";
+  private processExited: boolean = false;
 
   constructor(options: {
     command: string;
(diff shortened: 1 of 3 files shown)


packages/create-plugin/__e2e__/utils/CreatePlugin.ts
```diff
@@ -25,6 +25,7 @@ export class CreatePlugin {
   private currentStep: number = 0;
   private stdout: string = "";
   private stderr: string = "";
+  private processExited: boolean = false;
```
tuanphamcybozu, line 28: @hung-cybo 
prefix with is, has, can, or should.
These prefixes clearly indicate that the variable is a boolean and what it represents.

suggestion: `hasProcessExited`

packages/create-plugin/__e2e__/utils/CreatePlugin.ts
```diff
@@ -102,14 +112,29 @@ export class CreatePlugin {
...
+        this.processExited = true;
+        resolve();
+      });
+      this.finishProcess();
+    });
+  }
+
+  private finishProcess() {
     this.childProcess.stdin.end();
   }
```
tuanphamcybozu, line 139: @hung-cybo 
```suggestion
 private finishProcess(): void {
    if (this.childProcess.stdin) {
      this.childProcess.stdin.end();
    }
  }
```

Added a check to ensure `stdin` exists before calling end(). IMO, this prevents potential runtime errors if `stdin` is null or undefined.

tuanphamcybozu, line 139:
```suggestion
 private finishProcess(): void {
    if (this.childProcess.stdin) {
      this.childProcess.stdin.end();
    }
  }
```
Added a check to ensure stdin exists before calling end(). IMO, this prevents potential runtime errors if stdin is null or undefined.

tuanphamcybozu (approved):
LGTM

PR #2846, opened 2024-06-27
feat(rest-api-client): add space.addThread() method

shabaraba:
## Why

Support addThread method to be able to add a tread to a space.

## What

- Add new method addThread
- Add unit test
- Add demo script

## How to test

Check test spec

## Checklist

### DIFF
--- a/examples/rest-api-client-demo/src/space.ts
+++ b/examples/rest-api-client-demo/src/space.ts
@@ -83,6 +83,15 @@ export class Space {
     }
   }
 
+  public async addThread() {
+    const name = "Added Thread Name";
+    try {
+      console.log(await this.client.space.addThread({ space: SPACE_ID, name }));
+    } catch (error) {
+      console.log(error);
+    }
+  }
+
   public async updateThread() {
     const body = "<b>This is an updated thread body</b>";
     const name = "Updated Thread Name";
--- a/examples/rest-api-client-demo/src/space.ts
+++ b/examples/rest-api-client-demo/src/space.ts
@@ -84,7 +84,7 @@ export class Space {
   }
 
   public async addThread() {
-    const name = "Added Thread Name";
+    const name = "The thread added via rest-api-client";
     try {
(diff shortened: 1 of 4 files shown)


examples/rest-api-client-demo/src/space.ts
```diff
@@ -83,6 +83,15 @@ export class Space {
     }
   }
 
+  public async addThread() {
+    const name = "Added Thread Name";
```
tasshi-me, line 87: This would be easier to understand when checking on the kintone screen if it were changed as follows!

```suggestion
    const name = "The thread added via rest-api-client";
```

packages/rest-api-client/docs/space.md
```diff
@@ -175,6 +176,26 @@ An empty object.
...
+#### Parameters
+
+| Name  |       Type       | Required | Description                                                          |
+| ----- | :--------------: | :------: | -------------------------------------------------------------------- |
+| space | Number or String |   Yes    | The space ID.                                                        |
+| name  |      String      |   Yes    | The new name of the Thread.<br />Must be between 1 - 128 characters. |
+
+#### Returns
+
+An empty object.
```
tasshi-me, line 193: This method returns `id` as a response.
Please update this section 🙏

tasshi-me (changes requested):
I added two change requests.
Please check it 👍

examples/rest-api-client-demo/src/space.ts
shabaraba: addressed it in f0dceaac121df00fce20e8bbf53af50393ff9d9b

packages/rest-api-client/docs/space.md
shabaraba: addressed it in f0dceaac121df00fce20e8bbf53af50393ff9d9b

tasshi-me (approved):
LGTM!!!

hisasami (approved):
LGTM!

PR #2872, opened 2024-07-11
fix(rest-api-client): rename moveToSpace method to move

shabaraba:
## Why

The method name moveToSpace can be misleading.
In fact, it not only changes the space, but also removes the space.

## What

- change the `moveToSpace` method name to `move`

## Checklist


tasshi-me (approved):
LGTM!

PR #2950, opened 2024-08-30
feat(rest-api-client): Add method for uninstall plugin

shabaraba:
## Why

- Support plugin.uninstallPlugin method to be able to uninstalls a plug-in from the Kintone environment.

## What

- Add a document for the new method.
- Add new unit tests.

## How to test

1. prepare a zip file for a plug-in.
2. run below code and check logs:
```ts
import {KintoneRestAPIClient} from '@kintone/rest-api-client';

const client = new KintoneRestAPIClient();

// Upload a file and attach it to a record
const resp = await client.file.uploadFile({
  file: {
    path: "./files/plugin.zip"
  }
});

const fileKey = resp.fileKey;
console.log(fileKey);

if (typeof fileKey == "string") {
  const pluginId = (await client.plugin.installPlugin({fileKey: fileKey})).id;
  console.log(pluginId);
  console.log(await client.plugin.getPlugins({}));

  await client.plugin.uninstallPlugin({id: pluginId});
  console.log(await client.plugin.getPlugins({}));
}
```

## Checklist

### DIFF
--- a/packages/rest-api-client/docs/plugin.md
+++ b/packages/rest-api-client/docs/plugin.md
@@ -5,6 +5,7 @@
 - [getApps](#getApps)
 - [updatePlugin](#updatePlugin)
 - [installPlugin](#installPlugin)
+- [uninstallPlugin](#uninstallPlugin)
 
 ## Overview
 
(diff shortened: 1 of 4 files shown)


tasshi-me (approved):
LGTM!

PR #3016, opened 2024-10-10
fix(plugin-uploader): System ENV value check

charleslee0212:
env variable should not be empty.

## Why

The isSet function check to see if the env value is a type string; however, in the cli flags for the env are all defined as type string defeating the purpose of the isSet function. This results in not prompting the user to fill out the necessary values to run the kintone plugin uploader. The code will throw an error. This is a bug.

## What

I have updated the code to also check the value of the env. If the value is an empty string, isSet will return false.

## How to test

Run the kintone plugin uploader without initializing env variables. Then see if the code will prompt the user.

## Checklist

### DIFF
--- a/packages/plugin-uploader/src/params.ts
+++ b/packages/plugin-uploader/src/params.ts
@@ -7,9 +7,9 @@ import {
 } from "./prompts/params";
 
 interface Params {
-  username?: string;
-  password?: string;
-  baseUrl?: string;
+  username: string;
+  password: string;
+  baseUrl: string;
   lang: Lang;
 }
 
--- a/packages/plugin-uploader/src/params.ts
+++ b/packages/plugin-uploader/src/params.ts
@@ -13,7 +13,7 @@ interface Params {
   lang: Lang;
 }
 
-const isSet = (v: string | null | undefined) => typeof v === "string";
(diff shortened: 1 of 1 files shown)


chihiro-adachi:
@charleslee0212 
Thank you!
Please check the following two points.

- It looks like the CI is down, so could you check that?
- Is it possible to create a unit test for this fix?

chihiro-adachi:
プロンプトが動作することを確認。

```
$ unset KINTONE_BASE_URL
$ unset KINTONE_USERNAME
$ unset KINTONE_PASSWORD
$ bin/cli.js /path/to/plugin.zip
? kintoneのベースURLを入力してください (https://example.cybozu.com): http://localhost
? ログイン名を入力してください: username
? パスワードを入力してください: password
Open http://localhost/login?saml=off
Trying to log in...
Navigate to http://localhost/k/admin/system/plugin/
Trying to upload /path/to/plugin.zip
/path/to/plugin.zip をアップロードしました!
```

chihiro-adachi (approved):

chihiro-adachi:
@charleslee0212 
Thank you! it's merged.

charleslee0212:
Of course!

PR #3268, opened 2025-04-10
fix: make createClient callable without params

tasshi-me:
## Why

When used with session authentication, the parameter can be omitted.

## What

- make input parameter of createClient optional

## Checklist


shabaraba (approved):
LGTM

PR #3275, opened 2025-04-16
build(rest): enable SSR build to correct import target

tasshi-me:
## Why

The import/require target was incorrectly transformed on Vite output.

The output results are as follows.

```js
require(“/Users/tasshi/git/cybozu/js-sdk/node_modules/.pnpm/undici@7.4.0/node_modules/undici/index.js”)
```

- The local absolute path is specified.
- The entry point on the ESM side is specified

This is because Vite builds code for the browser.
(Browsers cannot import with the package name like `undici`, so Vite replaces it with the absolute path.)

By enabling SSR build, the build target is switched to Node.js, and the npm package name is retained.

```js
require(“undici”);
```

## What

- enable SSR build for ESM/CJS build

ref. https://vite.dev/config/build-options.html#build-ssr

## Checklist

### DIFF
--- a/packages/rest/package.json
+++ b/packages/rest/package.json
@@ -71,7 +71,6 @@
   },
   "devDependencies": {
     "vite": "^5.4.15",
-    "vite-plugin-cjs-interop": "^2.2.0",
     "vite-plugin-dts": "^4.5.3",
     "vite-plugin-license": "^0.0.1",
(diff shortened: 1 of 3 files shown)


chihiro-adachi (approved):

chihiro-adachi:
e2eが落ちていますが、https://github.com/kintone/js-sdk/pull/3277 で解消します。

PR #3537, opened 2025-12-18
fix(rest-api-client): set cursor size to 500 in getAllRecordsWithCursor

shabaraba:
## Why

refs: https://github.com/kintone/js-sdk-ja/issues/51

The `getAllRecordsWithCursor` method internally uses `createCursor` API, which has a default `size` of 100 records per request. However, the maximum allowed value is 500. By using the maximum size, we can reduce the number of API calls needed to fetch all records, improving performance.

For example, fetching 1000 records:
- With default size (100): 10 API calls
- With size 500: 2 API calls

## What

- Set `size: 500` when calling `createCursor` in `getAllRecordsWithCursor` method
- Updated the corresponding test to verify the size parameter is passed correctly

## How to test

1. Run the unit tests:
  ```bash
  cd packages/rest-api-client
  pnpm test -- src/client/__tests__/record/AllRecords.test.ts
  ```
2. Manual verification with a kintone app containing 1000+ records:
  ```ts
  const records = await client.record.getAllRecordsWithCursor({ app: appId });
  // Cursor should be created with size=500, reducing API calls
  ```

## Checklist

### DIFF
--- a/packages/rest-api-client/src/client/RecordClient.ts
+++ b/packages/rest-api-client/src/client/RecordClient.ts
@@ -316,7 +316,7 @@ export class RecordClient extends BaseClient {
   }): Promise<T[]> {
     const paramsWithSize = {
       ...params,
-      size: 500,
+      size: GET_RECORDS_LIMIT,
     };
     const { id } = await this.createCursor(paramsWithSize);
     try {


packages/rest-api-client/src/client/RecordClient.ts
```diff
@@ -314,7 +314,11 @@ export class RecordClient extends BaseClient {
     fields?: string[];
     query?: string;
   }): Promise<T[]> {
-    const { id } = await this.createCursor(params);
+    const paramsWithSize = {
+      ...params,
+      size: 500,
```
nameless-mc: https://github.com/kintone/js-sdk/blob/e482ecdbe143476c09df54e3d36e1f2348becbd0/packages/rest-api-client/src/client/RecordClient.ts#L19
を使う方が良さそう
```suggestion
      size: GET_RECORDS_LIMIT,
```

nameless-mc (approved):

PR #3697, opened 2026-04-02
fix: pin dependency versions and remove unused core-js

shabaraba:
## Why

The OSS packages currently published to the npm registry have their dependency versions specified with ^ in package.json. This notation locks only the major version and installs the latest compatible versions for minor and patch, meaning that when installing without a pnpm-lock.yaml (i.e., installing directly from the npm registry), there is a high risk of unintended package versions being installed.

Additionally, `core-js` was listed in `dependencies` for both `@kintone/rest-api-client` and `@kintone/rest`, but investigation revealed:

- **rest-api-client**: `core-js` is only used during the UMD build via Babel's `useBuiltIns: "usage"`. The `lib/` and `esm/` outputs (used by bundlers and Node.js) contain zero references to `core-js`. It is effectively a build-time dependency.
- **rest**: `core-js` is not referenced anywhere — not in source code, not in built output (`lib/`), and not even in the UMD bundle (which uses Vite's `target: "modules"` instead of Babel polyfills). It is completely unused.

## What

- Pin all `dependencies` versions across all packages (remove `^` prefixes)
- Add `package-json/restrict-dependency-ranges` ESLint rule to enforce pinned versions in `dependencies` going forward
- Exclude `templates/` package.json files from the pinning rule (end-user scaffolds)
- Move `core-js` from `dependencies` to `devDependencies` in `@kintone/rest-api-client`
- Remove `core-js` from `dependencies` in `@kintone/rest`
- Add `pnpm lint:package-json` command
- Update `renovate.json5`:
  - Set `rangeStrategy: "bump"` for `dependencies` to keep pinned versions up to date
  - Group related packages to update together (axios + form-data, openapi ecosystem, ajv ecosystem) to avoid version mismatch issues

## How to test

- `pnpm install && pnpm build` succeeds
- `pnpm --filter @kintone/rest-api-client test` — all 423 tests pass
- `pnpm --filter @kintone/rest test` — all tests pass
- `pnpm lint:package-json` passes with no errors
- UMD build sizes are unchanged:
  - `rest-api-client`: `KintoneRestAPIClient.min.js` = 245KB

## Checklist


packages/rest-api-client/package.json
```diff
@@ -63,32 +63,32 @@
...
-    "rollup-plugin-polyfill-node": "^0.13.0",
-    "vite": "^5.4.21",
-    "webpack": "^5.105.4",
-    "webpack-cli": "^6.0.1"
+    "@rollup/plugin-babel": "6.1.0",
+    "@rollup/plugin-commonjs": "26.0.3",
+    "@rollup/plugin-json": "6.1.0",
+    "@rollup/plugin-node-resolve": "16.0.3",
+    "@rollup/plugin-terser": "0.4.4",
+    "@types/core-js": "2.5.8",
```
nameless-mc: 多分不要

nameless-mc (commented):
packages/create-plugin/templates/**/package.json も対応してください

packages/rest-api-client/package.json
shabaraba: ありがとうございます、削除しました

chihiro-adachi (approved):

PR #3802, opened 2026-05-20
fix(plugin-manifest-validator): remove allowed_domains from Manifest v1

nameless-mc:
## Why

Follow-up to #3759, which introduced `sandbox` / `allowed_hosts` / `allowed_domains` / `permissions` on Manifest v1. The kintone backend decided to drop `allowedDomains` from the spec, so the validator side removes `allowed_domains` accordingly to keep the schema aligned with what kintone actually accepts.

## What

Remove `allowed_domains` from `@kintone/plugin-manifest-validator` Manifest v1:

- `manifest-schema.json`: drop the `allowed_domains` property definition.
- `manifest-schema.d.ts`: drop `allowed_domains` from the generated type.
- `src/__tests__/index.ts`: drop the `allowed_domains` test cases.

`sandbox` / `allowed_hosts` / `permissions` are out of scope and remain.

## How to test

- `pnpm --filter @kintone/plugin-manifest-validator test`
- `pnpm --filter @kintone/plugin-manifest-validator build`

## Checklist

### DIFF
--- a/packages/plugin-manifest-validator/manifest-schema.d.ts
+++ b/packages/plugin-manifest-validator/manifest-schema.d.ts
@@ -68,10 +68,6 @@ export interface KintonePluginManifestJson {
   };
   sandbox?: boolean;
   allowed_hosts?: string[];
-  /**
-   * Cross-domain access scope inside cybozu products. Defaults to "SELF".
-   */
-  allowed_domains?: "SELF" | "ANY";
   permissions?: {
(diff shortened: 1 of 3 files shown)


shabaraba (approved):

hisasami:
test spec in internal sharepoint (all cases passed): remove-plugin-allowed-domains.md

